PR #90, opened 2015-11-18
Removing obsolete obsoleteClass

### LINKED ISSUE #89: Error in parsing obo file
Hi,
I am trying to parse the latest [ro obo](http://purl.obolibrary.org/obo/ro.obo) file with [onto-perl](https://metacpan.org/release/ONTO-PERL) to into chado database. However, the parser is blowing up with the following error

```
The term with id 'ObsoleteClass' does NOT follow the ID convention: 'IDSPACE:UNIQUE_IDENTIFIER'
```

The offending line is [here](https://github.com/oborel/obo-relations/blob/master/ro.obo#L369).
Is this a issue with the ontology itself(obo version) or it is a problem with the perl module.

cmungall:
Fixes #89


cmungall:
Ha, in all the ontologies I work on we obsoleted ObsoleteClass a long time ago!

By messy, you just mean how it looks in Protege? I thought all ontology editors had gotten used to that ages ago. Seems wrong to force two redundant annotations, especially as one of those is enshrined in the OWL spec, and the other dates to an obsolete obo2owl translation.

Currently this messes up the obo translation, which uses the standard owl:deprecated tag
http://owlcollab.github.io/oboformat/doc/obo-syntax.html#5.8

I suppose we could add a hack to the obo translation for RO... but it still ends up as odd in our OWL import chains where we have eliminated ObsoleteClass.

Maybe a compromise is to put the obsolete class in an optional import?

jamesaoverton:
It looks messy in Protege, which I presume was the point in the first place. I'm not defending the practise.

http://purl.obolibrary.org/obo/BFO_0000005 was minted for ruttenberg-bfo2. It does not resolve. It was probably in `bfo-classes-minimal.owl` when I was cleaning things up, and I probably obsoleted it in this ugly OBI-ish way.

I dunno. It's not really RO's problem. Maybe we should remove both 'obsolete dependent continuant'  ObsoleteClass from `bfo-classes-minimal.owl`.

cmungall:
OK, let's do that. We should record somewhere the trail, but it doesn't need to be in the RO release

cmungall:
On RO call we agreed to get rid of ObsoleteClass and to merge this PR

balhoff:
See this Protege issue: https://github.com/protegeproject/protege/issues/270

There is a View menu option to hide deprecated classes from the hierarchy view.

matentzn (approved):
The diff is a bit spurious. I reviewed it and I am reasonably confident the only difference is the "subclassOf" obsolete class axiom.

For some reason the "entity refs" section  was removed, but not to the detriment of the file.

Issue #105, opened 2016-03-29
add 2 terms for IEDB & DO


rvita:
Hi, we would like to request 2 new terms to be used by the IEDB DO:
'has allergic trigger' is a relationship between a disease (domain) and material entity (range) where the realization of the disease results in pathological processes that includes an abnormally strong
immune response against the material entity which is not part of the host and is considered harmless to non allergic individuals.
Example: drug allergy (DOID_0060500) has_allergic_trigger some drug (CHEBI:23888).

'has autoimmune trigger' is a relationship between a disease (domain) and material entity (range) where the realization of the disease results in pathological processes that includes an immune response against the material entity, which is part of the host itself.

unsure of the parent terms b/c the RO_0002600 'capable of upregulating or causing pathological process' term causes a process, but a disease is a disposition.

elviram:
Hello,
any news regarding those two relations?

Thank you.

cmungall:
I'll take a look monday

jamesaoverton:
@cmungall I have a couple of changes to check with Bjoern and Randi, then I'll update the ticket.

jamesaoverton:
We want to use 'causes or contributes to condition' (http://purl.obolibrary.org/obo/RO_0003302) as the parent. So we'll swap the domain and range and adjust the definitions to better match that parent.

@cmungall Does this work for you? I can make the edits.

label: is allergic trigger for
parent: causes or contributes to condition
domain: material entity
definition: A relationship between a material entity and a condition (a phenotype or disease) of a host, in which the material entity is not part of the host, and is considered harmless to non-allergic hosts, and the condition results in pathological processes that include an abnormally strong immune response against the material entity.

label: is autoimmune trigger for
parent: causes or contributes to condition
domain: material entity
definition: A relationship between a material entity and a condition (a phenotype or disease) of a host, in which the material entity is part of the host itself, and the condition results in pathological processes that include an abnormally strong immune response against the material entity.

jamesaoverton:
@cmungall Are these OK with you? I can make the edits.

cmungall:
Yes these are fine. Note these don't represent the only way of modeling these; e.g. following a LEGO paradigm we would model this as the material entity being the substrate for some (possibly untyped) molecular event, with this having a causal relation to some downstream process, e.g. immune process

cmungall:
Aside regarding your drug example:

  * CHEBI:23888 is a role, not a material entity, so this would yield an 
unsatisfiable class
[...]

PR #112, opened 2016-06-30
First pass at inverse immune relations, #105

jamesaoverton:
It would be convenient to have the inverses of the two relations from #105.

Potential problem: As defined here, these forward and inverse versions use the same parent 'causes or contributes to condition'. Maybe the inverses need a different parent.


jamesaoverton:
Perhaps it makes sense to define an inverse for 'causes or contributes to condition' that could be the parent of these two relations.

PR #95, opened 2015-12-08
DO NOT MERGE YET First pass at #68 making a ro-chado subset

cmungall:
cc @cybersiddhu

Issue #68, opened 2015-07-23
Create documentation for use of RO in bioinformatics databases, particularly Chado


cmungall:
Discussion from list:

```
> So here's my very vague question: what do I need to know?  It seems that
> OBO_REL no longer exists, which is going to be a hassle for many
> Chado-related tools, which considered it fairly fundamental (with its is_a
> and part_of and a consistent cv.name of 'relationship').  I'm guessing BFO
> should completely replace OBO_REL, and RO is not really going to be needed
> in general instances of Chado (though obviously some users may want it, it
> isn't fundamental in the same way).

Hi Scott,

I recommend that Chado uses RO, you can get the .obo from here: http://purl.obolibrary.org/obo/ro.obo 

Note this includes some classes from BFO, it should be possible to produce a more minimal relation only edition/subset.

There are two pain points for you:

 1. Chado needs an edge label for "is_a"
 2. There is code that makes an assumption that there will be a relation with OBO-Format ID "OBO_REL:part_of" 

We can make a special edition just for Chado. For 1, we would add a special stanza with OBO-Format ID "OBO_REL:is_a". This doesn't really make any sense in terms of the mapping to OWL, where SubClass is a builtin construct. But we can ignore this, as this edition is just for Chado. 

For 2, it's up to you. You can try and coordinate the move for Chado to use BFO:0000050 (which is in RO). Or we can have the Chado edition have the old OBO_REL:part_of as well, with some documentation that the goal is to eventually move away from this.

```
cc @scottcain

scottcain:
Hi @cmungall 

I'm fine with creating a "Chado edition" of RO; the other thing we would certainly need is a 'derives_from', but it looks like that is present in RO, though I'm noticing that the names of all of the features in RO lack underscores--is that how they'd get stored in Chado as well?  Or is that part of what would be in the Chado edition?

Of course, it must also be loadable in Chado, and RO currently fails to load using xsltproc/DBIx::DBStag in Chado, but that may be due to the missing is_a or something else, I haven't looked closely at what the problem is.  If another method is preferable, that needs to be worked in to the build procedure.

cmungall:
This is also very important for people consuming the .obo version:

https://github.com/oborel/obo-relations/wiki/Identifiers

[...]

PR #115, opened 2016-08-12
Remove is_a from ro-chado.obo

abretaud:
As discussed in erasche/docker-tripal#6, we have a problem loading ro-chado.obo in chado.
It looks like removing "OBO_REL:is_a" from ro-chado fixes the problem.

### DIFF
--- a/subsets/ro-chado.obo
+++ b/subsets/ro-chado.obo
@@ -17,11 +17,6 @@ property_value: http://purl.org/dc/elements/1.1/title "OBO Relations Ontology" x
 property_value: http://xmlns.com/foaf/0.1/homepage " https://github.com/oborel/obo-relations/" xsd:anyURI
 
 [Typedef]
-id: OBO_REL:is_a
-name: is_a
-comment: This relation is added for backward compatibility with Chado
-
-[Typedef]
 id: BFO:0000050
 name: part_of
 def: "a core relation that holds between a part and its whole" []
--- a/subsets/ro-chado.obo
+++ b/subsets/ro-chado.obo
@@ -17,6 +17,11 @@ property_value: http://purl.org/dc/elements/1.1/title "OBO Relations Ontology" x
 property_value: http://xmlns.com/foaf/0.1/homepage " https://github.com/oborel/obo-relations/" xsd:anyURI
 
 [Typedef]
+id: OBO_REL:is_a
+name: is_a
+comment: This relation is added for backward compatibility with Chado
+
+[Typedef]
 id: BFO:0000050
 name: part_of
 def: "a core relation that holds between a part and its whole" []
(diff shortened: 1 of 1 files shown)


cmungall:
@cybersiddhu @kimrutherford comments? The fake is-a relation was added specifically to support chado, as discussed in #68

kimrutherford:
> @cybersiddhu @kimrutherford comments?

It won't be a problem for us.  We'll load is_a separately if needed.

scottcain:
I find it hard to believe that is_a is actually causing the problem @abretaud is describing.  I'll discuss over on their tripal tracker.

cybersiddhu:
Yes, the `is_a`  was a problem, as mentioned [here](https://github.com/oborel/obo-relations/pull/95#issuecomment-225553880), i have to use my custom fork to chop it out.
The reason was that the [perl obo parser](https://metacpan.org/release/ONTO-PERL) supplies its own `is_a` during parsing of any obo file, which conflicts with this explicit `is_a` in the `ro` file.

cmungall:
OK, we should move forward. The fix needs to be in the perl script though rather than derived file, see #119

abretaud:
Do you mean #119 fixed the problem or not? This can be closed if it's the case
Anyway, I'm working on a workaround in the script that loads obo into chado

cmungall:
I merged it, but also did #121 to ensure future releases don't include the fake isa

abretaud:
Great, thank you!

abretaud:
Hum, the fake is_a was reintroduced in this commit: fc42248de75e6b3c6ce6df6c25188271685e0725
Was it intended or not?

cmungall:
Yes it was intended. In fact ro-chado was originally created to provide 
this for some users, see for example 
https://github.com/oborel/obo-relations/issues/68#issuecomment-127390603

But those who need the fake stanza can add it separately, at it seems 
the majority want this excluded.

abretaud:
Ok, that's what I understood too, but the fake stanza is back in the latest revision

cmungall:
Oh, sorry I misunderstood. You are quite right. The last release contained an older version of ro-chado. The release Makefile now ensures that this is made as in sync with each release. I'll make a new release shortly

abretaud:
Hum, are you still thinking about the new release? (No hurry at all, just in case you forgot about it)

cmungall:
Permanent URL for this release: http://purl.obolibrary.org/obo/ro/releases/2016-09-02/subsets/ro-chado.obo

PURL for latest:
http://purl.obolibrary.org/obo/ro/subsets/ro-chado.obo

abretaud:
Awesome, thank you !
I just saw that there is now a `subsetdef: RO:0002259 ""` after the `"subsetdef: ro-eco ""` at the beginning, is it normal? I guess it has low impact, just wondering

cmungall:
It looks like the name of the slim/subset is not being persisted in the 
conversion from owl. This is an issue with the OWLAPI OWL->OBO (which I 
wrote..). It should be fixed, but in the interim this only affects you 
if you care about the RO subsets, and you want to be sure they have 
meaningful names. I don't think this is even included in a typical chado 
load.

abretaud:
Ok, yes, I think it should be ok like that
Thank you very much for your help!

PR #116, opened 2016-08-15
changing default namespace from "ro" to "relationship"; *many* pieces…

scottcain:
… of software expect relationship terms to be in a cv called "relationship"


scottcain:
@cmungall would you mind taking a look at this/commenting/accepting (whatever feels appropriate).  Thanks.

cmungall:
This is a derived file, so the change will either need to be in:

 * [translate-to-chado-obo.pl](https://github.com/oborel/obo-relations/blob/master/src/tools/translate-to-chado-obo.pl) (horrible hack, sorry)
 * the ro-edit.owl source

The latter takes a bit of understanding of the conversion process - if you need help file a ticket and I'll do it

Issue #130, opened 2016-10-27
Illegal redeclarations of entities: reuse of entity ... in punning not allowed


donpellegrino:
When making use of RO in Stanford Protege Desktop v5.1.0, a number of errors are reported to the log with the pattern "Illegal redeclarations of entities: reuse of entity ... in punning not allowed"

Level:       INFO
Time:        1477573771688
Message:     Finished loading imported ontology at http://purl.obolibrary.org/obo/ro/el-constraints.owl

Level:       ERROR
Time:        1477573771785
Message:     Illegal redeclarations of entities: reuse of entity http://purl.obolibrary.org/obo/BFO_0000063 in punning not allowed [Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/BFO_0000063>)), Declaration(AnnotationProperty(<http://purl.obolibrary.org/obo/BFO_0000063>))]

Level:       ERROR
Time:        1477573771786
Message:     Illegal redeclarations of entities: reuse of entity http://purl.obolibrary.org/obo/RO_0002222 in punning not allowed [Declaration(AnnotationProperty(<http://purl.obolibrary.org/obo/RO_0002222>)), Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/RO_0002222>))]

Level:       INFO
Time:        1477573771787
Message:     Finished loading imported ontology at http://purl.obolibrary.org/obo/ro.owl

Level:       ERROR
Time:        1477573771826
Message:     Illegal redeclarations of entities: reuse of entity http://www.w3.org/2004/02/skos/core#prefLabel in punning not allowed [Declaration(AnnotationProperty(<http://www.w3.org/2004/02/skos/core#prefLabel>)), Declaration(DataProperty(<http://www.w3.org/2004/02/skos/core#prefLabel>))]

Level:       ERROR
Time:        1477573771827
Message:     Illegal redeclarations of entities: reuse of entity http://purl.obolibrary.org/obo/BFO_0000063 in punning not allowed [Declaration(ObjectProperty(<http://purl.obolibrary.org/obo/BFO_0000063>)), Declaration(AnnotationProperty(<http://purl.obolibrary.org/obo/BFO_0000063>))]

Level:       ERROR
Time:        1477573771827
[...]

PR #131, opened 2016-10-27
Fixed AnnotationProperty/ObjectProperty illegal punning

balhoff:
Fix for #130.


balhoff:
Actually I think I missed one (prefLabel). Will need to look into that later.

PR #142, opened 2017-01-13
added new terms

diatomsRcool:
I added some terms I need for my globi files. The changes are in ro-edit.owl


cmungall:
Thanks!

I don't think the build failure is due to your changes, something that happened before

PR #145, opened 2017-01-31
Moved problematic property chains to SWRL rules.

balhoff:
In response to the discussion on 1bb0e22d1a554addd652e02eaa3e45fd77a78b64, I created three SWRL rules:

```
enables(?x, ?y), 'occurs in'(?y, ?z) -> 'part of'(?x, ?z)
'negatively regulates'(?x, ?y), 'positively regulates'(?y, ?z) -> 'negatively regulates'(?x, ?z)
'positively regulates'(?x, ?y), 'negatively regulates'(?y, ?z) -> 'negatively regulates'(?x, ?z)
```

This should make the file valid OWL DL. I also removed the property chain version of these.

### DIFF
--- a/src/ontology/ro-edit.owl
+++ b/src/ontology/ro-edit.owl
@@ -1,16 +1,18 @@
 <?xml version="1.0"?>
 <rdf:RDF xmlns="http://purl.obolibrary.org/obo/ro.owl#"
      xml:base="http://purl.obolibrary.org/obo/ro.owl"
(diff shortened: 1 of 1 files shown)


balhoff:
@cmungall @dosumis what do you think about having SWRL in the oborel release? I don't have privileges here to merge this.

cmungall:
We'll want to explore implications for modularization strategy, but these are valid and syntactically correct DL so I merged.

PR #160, opened 2017-05-10
New, more readable relation definitions for regulates relations in RO

dosumis:
I got sick of trying to justify the old, frankly unreadable definitions on the regulates relations.  This pull request replaces these old definitions with ones I presented to GO in LA in November 2016.  I just reviewed them with the causal relations working group and all agreed to the change.  Note that we deliberately removed the clause 'x is the realization of a function to exert an effect on' and have not replaced it with anything equivalent.  IIRC, this clause was meant to restrict use of the relation to  processes that realize canonical, 'evolved' regulatory functions.   We think that a more general relation is needed for integration purposes and that such restrictions should be imposed when the relations are used, if (as in GO curation or the GO ontology) it is appropriate to do so.

### DIFF
--- a/src/ontology/ro-edit.owl
+++ b/src/ontology/ro-edit.owl
@@ -26,8 +26,8 @@
         <owl:imports rdf:resource="http://purl.obolibrary.org/obo/ro/go-biotic.owl"/>
         <owl:imports rdf:resource="http://purl.obolibrary.org/obo/ro/el-constraints.owl"/>
         <dc:description xml:lang="en">The OBO Relations Ontology (RO) is a collection of OWL relations (ObjectProperties) intended for use across a wide variety of biological ontologies.</dc:description>
-        <foaf:homepage rdf:datatype="http://www.w3.org/2001/XMLSchema#anyURI"> https://github.com/oborel/obo-relations/</foaf:homepage>
         <rdfs:comment rdf:datatype="http://www.w3.org/2001/XMLSchema#string">Includes Ontology(OntologyID(Anonymous-21)) [Axioms: 26 Logical Axioms: 0]</rdfs:comment>
+        <foaf:homepage rdf:datatype="http://www.w3.org/2001/XMLSchema#anyURI"> https://github.com/oborel/obo-relations/</foaf:homepage>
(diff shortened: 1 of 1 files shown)


cmungall:
Do you know what happened here, you introduced this:

https://github.com/oborel/obo-relations/blob/70aa2792ff6e0a58ee2f6868ea533199aae5d23c/src/ontology/ro-edit.owl#L49-L53

This may have been latent already and the OWLAPI materialized it... but we can't have APs being subprops of OPs

cmungall:
+1 to the def changes. We can merge and then fix the AP error later up to you

dosumis:
No idea how the AP error happened.  I just opened in Protege 5.2 and edited.  Rather worrying.   The file wasn't showing up as edited prior to this session.

dosumis:
Hi Chris,

I can’t see a clean way to fix in Protege as I need to delete AP entity declarations for these object properties and that doesn’t appear to be possible in Protege without side effects.  Probably best to delete declarations by hand from functional syntax file.  Do you agree? 

As to where this came from:  I’m wondering whether it might be the result fo OWL-API interpretation of scripty edits to the ontology?   Have these APs that link OP to OP been in place long?  (e.g. is direct form of).  Could these be confusing Protege?

Cheers,
David

I’m wondering whether this is a downstream result

cmungall:
On 10 May 2017, at 13:07, David Osumi-Sutherland wrote:

> I can’t see a clean way to fix in Protege as I need to delete AP 
> entity declarations for these object properties and that doesn’t 
> appear to be possible in Protege without side effects. Probably best 
> to delete declarations by hand from functional syntax file. Do you 
> agree? 

+1

> As to where this came from: I’m wondering whether it might be the 
> result fo OWL-API interpretation of scripty edits to the ontology?

Possible.

> Have these APs that link OP to OP been in place long? (e.g. is direct 
> form of). Could these be confusing Protege?

They've been there a while. Should be totally valid. It's possible that 
somehow Protege is munging.

dosumis:
OK. Should now be fixed.  Is there any reason not to have the master in functional for diffs?  Or should I be hooking in Shahim's OWL diff tool?

cmungall:
I use shahim's on the command line

RDF/XML because history. Could do with general spring cleaning - changing all source files to functional

PR #162, opened 2017-05-24
First draft of reg component relations for GO MF refactoring

dosumis:
image
image

Decided to go with more specific relations to avoid confusion in RO context with continuant-continuant relations  e.g. Could a thermostat be considered a regulatory component of a heating system?

Happy to go with simpler alternative if you'd prefer though.  Could make intention clear with domain and range + comments.  

Note - i've used this branch to create  a temporary file of relations for MF refactoring

https://github.com/geneontology/go-ontology/blob/MF_refactoring/src/ontology/component_activity_relations_tmp.owl

So would like to keep IDs stable.

### DIFF
--- a/src/ontology/ro-edit.owl
+++ b/src/ontology/ro-edit.owl
@@ -669,6 +669,18 @@ Where we have an annotation assertion
     
 
 
+    <!-- http://www.geneontology.org/formats/oboInOwl#created_by -->
+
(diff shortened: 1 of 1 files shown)


dosumis:
>  I feel we may need a more detailed account of necessity, but this can be done in the future.

You're probably right.  But may help to test it against real examples first.

PR #165, opened 2017-05-26
Food consumer relations

Public-Health-Bioinformatics:
Adding relations: 
- has consumer
- immersed in
- has substance added
- has primary substance added

Primarily for FoodOn food ontology.


Public-Health-Bioinformatics:
Ah, dear.  So we need to toy with 'has potential consumer' or 'capable of being consumed by' as a name.  Ok, thanks for taking this in for now… I can make use of the relation Ids right away.

Historicall, there was an approach of defining "[entity] 'has role' some [role]", and


D.

cmungall:
Not so different, we'd define as a shortcut of:

C SubClassOf has-role some (realized-by some consumption and has-input some D)

GO:0007631 may be appropriate for the 'consumption' process here.

PR #175, opened 2017-06-06
Fix typos in exposure relation definitions

jamesaoverton:


jamesaoverton:
I believe that the definition of 'has exposure receptor' should now refer to 'exposure stimulus' rather than 'exposure stressor', but I might have misunderstood the last batch of changes.

There was a typo with 'recaptor'.

cmungall:
Thanks. I thought I got the recaptor but obviously not

PR #213, opened 2017-09-30
agonist and antagonist relations for GO/GOCAM

dosumis:


dosumis:
Decided not to go with this one.

PR #220, opened 2017-12-07
Added in_reg_with relation for VFB.

dosumis:


dosumis:
We use FOAF:image and FOAF:depicts in VFB.  This originated with requirements for Jim's image plugin.  Potentially useful to have domain and range on this relation.

IAO:0000101 doesn't quite work because it specifies 2D.
&
FBbi:00000266 doesn't work because its an imaging method and anyway specifically excludes recorded images.

I could move the domain and range into VFB_ext.owl.  Or perhaps just sufficien to move FOAF typing under BFO to there?

cmungall:
perhaps move to VFB for now?

We can separately engage IAO to see if image can be redefined.

dosumis:
Open IAO issue: https://github.com/information-artifact-ontology/IAO/issues/203

Might be worth some concerted effort to define a set of compositional image terms using the imaging method hierarchy from FBbi - which seems to be the best thought out and most intuitive attempt at this.

PR #225, opened 2018-01-26
Berkeley go mtg jan 2018

cmungall:


cmungall:
@vanaukenk was this everything that was required? I don't think we made a main ticket

vanaukenk:
@cmungall - is there anything more we need to do here for this pull request?  
I see we have the positive and negative effect children of the GP-BP relations and the enables_subfunction relation.  Is there anything more we need to do for the is_active_in relation?  
Thx.

PR #231, opened 2018-05-21
First pass at atomic/componentized release

cmungall:
https://github.com/INCATools/ontology-starter-kit/issues/50

cc @balhoff


balhoff:
@cmungall I was going to update this to use `-base` suffix, but there's quite a lot of other changes here.

PR #234, opened 2018-06-25
Update ro-idranges.owl

cmungall:
Adding id range for @msinclair2

See #233


msinclair2:
What type of property chain exactly?  One that specifies that generically depends on is a subproperty of { RO:concretized_by, RO:specifically_depends_on }?

PR #236, opened 2018-07-08
Obsoleted "related via dependence to"

### LINKED ISSUE #235: treatment of regulates relations and their parents
The GO regulates relations are part of a hierarchy, illustrated here:
+ related via dependence to
   + causally related to
       + causal relation between processes
              + causally upstream of or within
                 + causally upstream of
                     + causally upstream of, positive effect
                     + causally upstream of, negative effect
                     + immediately causally upstream of
                     + transitively provides input for
                     - regulates (processual)
                           - has effector activity
                           + negatively regulates
                           + positively regulates
                           + directly regulates
 
 
An example is as follows: 
ObjectProperty: related via dependence to
Term IRI: http://purl.obolibrary.org/obo/RO_0002609
Definition: A relationship that holds between two entities, where the relationship holds based on the presence or absence of statistical dependence relationship. The entities may be statistical variables, or they may be other kinds of entities such as diseases, chemical entities or processes.
There would seem to be a number of problems with this definition, including:
 
1. it defines 'dependence relationship' in terms of 'statistical dependence relationship' which looks circular
2. it allows a relationship of dependence to hold in virtue of the absence of a statistical dependence relationship, which seems odd.
3. what, from the OBO Foundry perspective, is a statistical variable? is it an entity?

cmungall:
Fixes #235


cmungall:
Agh, ignore, this, accidental merge commit

PR #256, opened 2018-09-12
Adding domain-range constraints for located_in, see #255

cmungall:
note this changes introduces a spurious diff bringing in
AP declarations, this can be ignored


alanruttenberg:
Hi Chris,

Barry asked that I have a quick look at this. 

In the new axiomatization, located_in is (independent continuant other than spatial region) -> (independent continuant other than spatial region).

For relating to spatial regions, the relation is occupies-spatial-region and is (independent-continuant other than spatial region) ->  spatial region. 

There is no location relation from spatial region to spatial region. In that case part-of is the appropriate relation.

cmungall:
Thanks - I'm taking care of these on this PR: #261

PR #261, opened 2018-09-26
Ready to merge: Domain and range constraints

cmungall:
- Adds multiple new domain and range constraints
 - Brings in more external classes (still need to refactor import generation)
 - Adding more obo foundry unique labels
 - Adding additional metadata / AAs


cmungall:
I don't think the GO orphan issues were introduced with this PR, but this is a good time to fix this.

robot: I've always used the robot on the path. We should switch the .travis.yml to use the odk docker container

Issue #164, opened 2017-05-24
Define design patterns for relational constraints


cmungall:
For example, a `homeomorphic` DP that takes args R, C

 * C SubClassOf R only C

Or `never-between`:

 * C DisjointWith R some D
 * D DisjointWith R some C

#163 is an example of this

nlharris:
Looks like maybe this is done?

nlharris:
See also #52.

Can this ticket be closed?

PR #270, opened 2018-10-21
Ready to merge: homemorphy meta-relations see #164

cmungall:
Added homeomorphy declarations for part-of and in-similarity-relation-with, see #164


cmungall:
TODO: add SPARQL to generate axioms from meta-relation shortcut

PR #276, opened 2018-10-23
Added SWRL rules from domain/range specific relations

### LINKED ISSUE #275: Add SWRL rules to core.owl for inference of domain/range specific relations
For context see: https://github.com/oborel/obo-relations/wiki/domain-or-range-specific-relation

cmungall:
fixes #275


cmungall:
Yes it is

We should add c part of. How far deep though?

nlharris:
Can this be merged?

PR #300, opened 2019-01-31
Make causally_upstream_of and causally_downstream_of disjoint with overlaps.

balhoff:
If x is causally upstream of y, they should not share parts.


balhoff:
It turns out non-simple properties should not be used in DisjointObjectProperties axioms. I will convert this to a SWRL rule.

PR #322, opened 2019-05-07
removed range constraint on has-input

### LINKED ISSUE #321: Add warning to non-temporally quantified relations
Some relations need a comment clarifying their temporal interpretation and warning about uses which are potentially unsafe. 

Draft for part_of (BFO:0000050):

part_of is potentially unsafe to use between two classes of continuant where one or both are non-rigid or where a relationship is transitory.  This is particularly a problem when building models of developmental anatomy or recording the relationships of migrating cells to tissues.

Examples:  
A cell type is part_of a primordium at an early stage and part_of a mature structure at a later stage.  In this case BFO:000050 should NOT be used to link the cell type to either structure.

.... (MORE WORK NEEDED)

cmungall:
fixes #321


cmungall:
Note there are spurious diffs here

This is the only semantic change:
https://github.com/oborel/obo-relations/pull/322/files#diff-42218b151d03bc8af4184265cb158d79L2381

It removes the material entity range constraint on has-input.

cmungall:
Yes, issue 315.

I'm not totally following your last paragraph. Are you saying you don't care one way or another about the strictness of the range of 'has input'? But wouldn't we want OBI's 'has specified input' to be a subproperty of 'has input'?

jamesaoverton:
All I'm saying is that OBI's 'has specified input' has never been a subproperty of 'has input', and I was not pushing to change that situation.

cmungall:
> All I'm saying is that OBI's 'has specified input' has never been a subproperty of 'has input', and I was not pushing to change that situation

But there is no reason not to make this change after this PR is merged? (provided there is some kind of SLA between RO and OBI that we won't revert and break things). This would restore a certain amount of terminological harmony to the universe.

There is a broader question here of aggregating relations in RO vs allowing ontologies autonomy, and I get there are challenges with both

matentzn:
We will revisit this when COB owns planned process

jamesaoverton:
I did a poor job of explaining this to @matentzn: As far as I'm concerned, RO can relax the range of 'has input' now (or any time).

I would rather roll any changes to 'has specified input' into our COB work, but that's a downstream question.

matentzn:
I removed the conflicts, and now it should be clear. I will merge it.

matentzn (approved):

PR #324, opened 2019-05-15
Fix wikipedia URL in `rohom.owl`

althonos:
`rohom.owl` contains an invalid URL to a Wikipedia cross-reference, that ends up in `ro.obo` when building the ontology.


cmungall:
This is generated from the `hom` ontology, will fix here for now since we don't regen that often

althonos:
I'll open a PR to [BgeeDB/homology-ontology](https://github.com/BgeeDB/homology-ontology) as well.

althonos:
@cmungall: current release still have the issue, do you think you could regenerate `ro.obo` ?

Issue #328, opened 2019-06-20
Continuant part-of should be declared transitive


cmungall:
More generally, we should have a DP for any homeomorphic subrelation of a transitive relation (e.g. X part of where X is both domain and range) that this is transitive

wdduncan:
This issue is being closed because it has not had any activity in the last 2 years. If you think it is still worth pursuing, please reopen the issue and comment on its status.

PR #329, opened 2019-06-20
Adding molecular-part-of. See aslo #328

cmungall:
Making active-in subproperty of located-in


matentzn:
PArt of the bigger issue to solve the specific vs generic relations. DIscuss in that context.

PR #339, opened 2019-10-17
Edits to synapsing relationships

### LINKED ISSUE #286: in neural circuit with Domain is too strict
With the new RO we are getting this for Uberon:

image

Clare72:
Moved synaped_by and synaped_to out from under in_neural_circuit_with and moved electrically_synapsed_to under in_neural_circuit_with
Edit to definition of downstream_in_neural_circuit_with (was the same as upstream def!)
Made upstream and downstream terms inverses
Fixes #286


Clare72:
Definitions of in_neural_circuit_with, downstream_in_neural_circuit_with, upstream_in_neural_circuit_with and electrically_synapsed_to say they are between two neurons.

Unless you want to change the definitions, synapsed_to and synapsed_by shouldn't be under there at all.

dosumis:
Good point.  If we keep the relations,  textual definitions should just specify cells, not neurons.  Moving the relations out is not a good solution, as the relations then become useless - as there will be no inference to them.

Clare72:
Is it worth making new relationships for this?

'connected via neural circuit to'
'downstream via neural circuit of'
'upstream via neural circuit of'

We could keep the 'in neural circuit' relationships to connect only neurons (keeping existing definitions)

cmungall:
Hi @Clare72 - it looks like we have a conflict. I don't know how familiar you are with functional syntax but if you click on Resolve Conflicts you will be given options as to which axiom to choose.

dosumis:
We discussed this on the RO call - 30th June and decided that ` in_neural_circuit_with` should be obsoleted.

Clare72:
and upstream/downstream children?
shall I put synapsed_to/by under sends_output_to/receives_input_from?
and electrically_synapsed_to under connected_to?

Clare72:
obsoleted neural circuit relationships
electrically_synapsed_to now SC of connected_to
synapsed_to now SC of sends_output_to
synapsed_by now SC of receives_input_from

PR #348, opened 2020-02-05
Fixes #347, but do we want to?!

### LINKED ISSUE #347: 'has effector activity' should not be a subclass of 'regulates'
Hi @cmungall 

has effector activity http://purl.obolibrary.org/obo/RO_0002025 
should not be a subclass of 'regulates' 

It is 'downstream' of the regulatory activity: 

image

Thanks, Pascale

dosumis:
See ticket  for discussion


dosumis:
@pgaudet @cmungall

cmungall:
Will make a new PR using a chain instead, see ticket

Issue #350, opened 2020-02-14
Add  causal relations between phenotypes


dosumis:
We already have  Process(P) -> Process causal relations.  At this point these are considered atomic. Derived from these are Continuant (C) -> P, C->C (via P) and P->C.

![image](https://user-images.githubusercontent.com/112839/74526838-2bc17a00-4f1c-11ea-8af8-71d533a4b11e.png)

We have multiple use-cases for making causal links between phenotypes/qualities (Q->Q) but no relations so far 

e.g.  see https://github.com/obophenotype/upheno/issues/590

P -> Q is also needed by  GO (who use the regulates relation illegally for regulation of quality, in the absence of something better)

Good axiomatization wiring these new relations up to primitive causal relations between processes is challenging, but (IMO) should not hold-up adding urgently needed relations.

cmungall:
I suppose the reason incoherency has not shown up for GO RoQs is the fact we have not rolled out Elk 0.5 yet (cc @balhoff).

I worry about the proliferation of parallel hierarchies for all 9 combinations of `{P,ME,Q} x {P,ME,Q}`.

How would we feel about grouping Q and P for the purposes of domain/range constraints?

PR #351, opened 2020-02-14
First draft attempt at some Q2Q causal relations.  See #350

dosumis:
DON'T MERGE YET!  More discussion needed.


dosumis:
Currently going with phenotype_results_in and phenotype_results_from, maybe simplify to phenotype results_from  / results_in ?

There is demand for weaker sounding relations, do we add these?

Some examples (hemolytic anemia) might better be  considered P->Q not Q->Q

For use within a phenotype def, should we restrict direction to results_from?

Ultimate aim is to define with reference to process -> process causal relations if poss but this will take work.

matentzn:
Does the label "phenotype" capture the wild type (i.e. normal phenotypes) as well? If so, I think I am good with this.

PR #353, opened 2020-02-26
Removing Obsolete Domain from 'has ligand'

### LINKED ISSUE #306: Domain of obo:RO_0002019 (has ligand) is obsolete
Currently:
[GO_0004872](http://www.ontobee.org/ontology/GO?iri=http://purl.obolibrary.org/obo/GO_0004872)
which is now obsolete. Term replaced by [GO_0038023](https://www.ebi.ac.uk/ols/ontologies/go/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FGO_0038023)

Shows up in ontologies importing RO here and there.

matentzn:
Fixes #306

You need to decide whether you want the Domain to be replaced by GO:0038023 or not, but this axiom should go.


pgaudet:
Hi @matentzn 

We are planning to create a child of GO:0038023, 'small molecule-activated receptor activity'. Can you hold off and replace the domain once that new term is available ?

matentzn:
Of course :) I would never dare to merge a PR on RO, so this is in any case for @cmungall to do; who I am sure will gladly wait.

bpeters42:
This label seems wrong to me. has_ligand has a textual definition that matches my intuitive understanding that this is referring to a relationship between a receptor and a ligand, two material entities. But the domain here is a process. Something like 'has ligand participant' is a better label.

cmungall:
On RO call we agreed to merge this as the axiom is no longer valid.

There are also general concerns about this relation, but this should be dealt with in a separate ticket, and will be discussed on RO call about has-input

PR #364, opened 2020-04-06
Update ro-base.owl

jhammock:
added definition for http://purl.obolibrary.org/obo/RO_0002454, "has host":  "X 'has host' y if and only if: x is an organism, y is an organism, and x can live on the surface of or within the body of y"


balhoff:
Hi @jhammock could you please redo your PR on this file: https://github.com/oborel/obo-relations/blob/master/src/ontology/ro-edit.owl

The one you edited is generated at release time. Some of this info should be in CONTRIBUTING.md. I filed an issue about getting that more up to date.

balhoff:
By the way the edit file is OWL Functional syntax rather than RDF-XML.

PR #385, opened 2020-08-06
Changing expansion of capable of, wrong relation

matentzn:
As far as I can tell, there is NO RO_0000054. There is, however, BFO_0000054 (realizes in), so I assume that this is what this expansion was intended to be.


cmungall (approved):
Thanks

We should move away from use of IAO_0000424

PR #386, opened 2020-08-13
Relations 384

pgaudet:


cmungall:
I think this should be added as a synonym for:

```
id: RO:0002491
name: existence starts and ends during
def: "x existence starts and ends during y if and only if the start of x is part of y and the end of x is part of y. Formally: x existence starts and ends during y iff α(x) >= α(y) & α(x) <= ω(y) & ω(x) <= ω(y) & ω(x) >= α(y)" []
is_a: RO:0002488 ! existence starts during
is_a: RO:0002492 ! existence ends during
```

pgaudet:
Ahhh ! so you dont think this should be a new term ? 
@vanaukenk

vanaukenk:
@pgaudet @cmungall @ukemi

Okay, I thought we'd approved this new term and its placement in RO on the 2020-08-10 ontology call, but if not, then I want to make sure I'm clear on the definition of 'existence starts and ends during'.

Wrt this part of the definition:  

"the start of x is part of y and the end of x is part of y"

Should this be interpreted to mean that there is a process that corresponds to the 'start of x' that is part of 'y' where y is a GO BP or some developmental stage?

If so, then I'm a little confused about the parentage, i.e. 'relation between physical entity and a process or stage'.

ukemi:
I have the same questions as @vanaukenk. What is the start and end of a continuant? I think those were our original concerns with this relation.

PR #410, opened 2020-11-24
Ro issue 391

pgaudet:


src/ontology/ro-edit.owl
```diff
@@ -1005,6 +1009,34 @@ AnnotationAssertion(rdfs:label obo:RO_0040042 "is homeomorphic for")
 AnnotationAssertion(rdfs:seeAlso obo:RO_0040042 <https://github.com/oborel/obo-relations/issues/164>)
 SubAnnotationPropertyOf(obo:RO_0040042 obo:RO_0002422)
 
+# Annotation Property: obo:valid_for_go_annotation_extension (valid_for_go_annotation_extension)
```
cmungall, line 1012: unfortunately subsets are fiddly in obo/owl, this is in a global obo namespace, should be in an RO one. I'll fix with a search and replace

PR #413, opened 2020-12-04
Add relation chain 'process has causal agent' o 'in taxon' -> 'in taxon'

### LINKED ISSUE #409: 'process has causal agent' : add property chain for 'in taxon'

balhoff:
Fixes #409.

### DIFF
--- a/src/ontology/ro-edit.owl
+++ b/src/ontology/ro-edit.owl
@@ -6267,6 +6267,7 @@ SubObjectPropertyOf(ObjectPropertyChain(obo:RO_0002583 obo:BFO_0000050) obo:RO_0
 SubObjectPropertyOf(ObjectPropertyChain(obo:RO_0002583 obo:BFO_0000062 ObjectInverseOf(obo:RO_0002583)) obo:RO_0002496)
 SubObjectPropertyOf(ObjectPropertyChain(obo:RO_0002593 obo:BFO_0000050) obo:RO_0002492)
 SubObjectPropertyOf(ObjectPropertyChain(obo:RO_0002593 obo:BFO_0000063 ObjectInverseOf(obo:RO_0002593)) obo:RO_0002497)
+SubObjectPropertyOf(ObjectPropertyChain(obo:RO_0002608 obo:RO_0002162) obo:RO_0002162)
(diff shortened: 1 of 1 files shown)


wdduncan:
Hi @balhoff.  
Sorry, if I'm coming late to the party :)  
Can you clarify what the domain and range are of `process has causal agent` are?  The definition in `ro-edit.owl` seems (to me) to indicate that the domain is a process and the range is material entity. If so we have:

Let **p** be a process, **m** be a material entity, and **t.** the target taxon Then I think the chain role would imply:

**p** `process has causal` **m** `in taxon` ->  **p** `in taxon` **t**

Are you wanting to say that process **p** is in taxon **t**?

Apologies if I am not understanding the intent or logic involved.

balhoff:
Hi @wdduncan, this change is to address a specific issue in GO: https://github.com/geneontology/go-ontology/issues/17613#issuecomment-720491074

We do want to infer that the process p is in taxon t. Since this came from a GO requirement, I wanted to make sure a couple other people reviewed to catch any cases where this might not hold generally.

dosumis:
Does GO still have multi-organism processes that are multi-species (e.g. for symbiosis)?  Inference may not be safe in these cases.

wdduncan:
thanks @balhoff 

The definition for `in taxon` reads:
```
x is in taxon y if an only if y is an organism, and the relationship between x and y is one of: part of (reflexive), developmentally preceded by, derives from, secreted by, expressed.
```
If processes can be the the domain the `in taxon` will you need to update the definition?

balhoff:
> If processes can be the the domain the in taxon will you need to update the definition?

I didn't realize that was in the 'in taxon' definition. Processes are already widely used in the domain of 'in taxon', so I think that's a separate issue to address.

dosumis (changes requested):
May not be safe given that some processes may have causal agents in multiple species - e.g.- root nodule formation.  This may already apply to terms in GO.

balhoff:
Thanks @dosumis! We'll discuss at a GO call.

balhoff:
We agreed at the GO call on 2020-12-14 that this is not general, so I'll close this.

Issue #419, opened 2021-01-12
new term: "co-roosting with", subtype of "ecologically co-occuring with"


jhpoelen:
In recent discussions (see https://github.com/globalbioticinteractions/globalbioticinteractions/issues/583), it appears that co-roosting is an ecologically relevant interaction term especially for bats (Chiroptera). Suggest to add the co-roosting term as a "child" of the term ecologically co-occurs with.

jhpoelen:
@balhoff @cmungall what is the most efficient way to introduce a term these days?

balhoff:
I think PRs are very welcome, but the need to generate a term ID complicates things. I'm not set up for that. Do you have an ID range assigned to you? Or maybe @diatomsRcool does?

jhpoelen:
@balhoff Thanks for your response!

As far as I know, I don't have an ID range assigned to me. What is the process for getting an ID range assigned? Happy to put in some work so that I do not have take time from others to create a pull request. However, as I am getting up to speed, I think I'd benefit from looking over the shoulder of a more experience contributors (like @diatomsRcool) as they prepare an example pull request for the proposed co-roosting RO term.

jhpoelen:
fyi @qgroom

balhoff:
I have a dream of eliminating the need for ID ranges :-) (https://github.com/ontodev/robot/issues/763). I'll probably implement that soon but it will be a while before getting into workflows. Hoping @cmungall can comment on handling ID ranges in the immediate term. Here are the folks who are set up to mint new IDs: https://github.com/oborel/obo-relations/blob/master/src/ontology/ro-idranges.owl

cmungall:
go ahead and make a PR on ro-idranges.owl and just assign yourself an
unused block

qgroom:
Here is a suggestion for a definition, please feel free to correct if you don't think this covers it.

The propensity for different species to actively congregate in the same place for rest or sleep

jhpoelen:
An updated definition by Nancy Simmons - 

Bat Co-roosting:  Two or more bats (Order Chiroptera) sharing the same roost site (cave, mine, tree hollow, animal burrow, leaf tent, rock crack, space in man-made structure, etc.).

Note that this is different from birds co-roosting.

cmungall:
Hmm, my immediate reaction is that this is excessive pre-composition, we could end up with subproperties named for many different taxa.

[...]

PR #421, opened 2021-01-16
add unused id range for Jorrit Poelen as suggested by @cmungall

jhpoelen:
related to https://github.com/oborel/obo-relations/issues/419#issuecomment-761122111


jhpoelen:
@cmungall thanks!

PR #433, opened 2021-03-01
Remove property chains inferring transports, imports, exports for all transitive parts of the thing being transported.

balhoff:
Result from discussion at https://github.com/geneontology/go-ontology/issues/20872

We (GO) decided that the notion of, e.g., "exporting" something (a particular compound) doesn't carry down to all the component parts of that thing (like each hydrogen atom). The hydrogens are moved but not specifically "exported".

I consider using "results in movement of" as the property chain result, but this property seems anatomy-specific for some reason.


balhoff:
If no one objects, I'll merge this early next week.

PR #435, opened 2021-03-08
RO Release March 2021

matentzn:


ro-base.obo
```diff
@@ -5408,6 +5413,16 @@ is_a: RO:0040036 ! has anatomical participant
...
 creation_date: 2020-08-27T08:13:59Z
 
+[Typedef]
+id: RO:0012008
+name: results in fusion of
+subset: http://purl.obolibrary.org/obo/valid_for_go_annotation_extension
+subset: http://purl.obolibrary.org/obo/valid_for_go_ontology
+subset: http://purl.obolibrary.org/obo/valid_for_gocam
+is_a: RO:0002592 ! results in organization of
+created_by: pg
```
cthoyt, line 5423: can this field take an orcid ID instead?

ro-base.obo
```diff
@@ -5408,6 +5413,16 @@ is_a: RO:0040036 ! has anatomical participant
 created_by: pg
 creation_date: 2020-08-27T08:13:59Z
 
+[Typedef]
```
cthoyt, line 5416: would be nice to enforce that there's a definition for this typedef

balhoff (approved):

ro-base.obo
matentzn: It could, but there is a bit of inconsistency right now in the world of `created_by`.. Not an issue for this release, but we should standardise on OBO level.

ro-base.obo
matentzn: I agreee. Make a ticket. Oddds are someone will fix it soon (during the RO meeting).

cthoyt: hmm in addition, the CI should be checking to make sure that all fields have this, and then you can simply reject any commit that doesn't pass CI (it's much better to have curators get mad at the computer for having very strict standards than you as the arbiter of the repo)

matentzn: Yeah we usually do that, but since there are now so many typedefs without definitions, we would have to do some awkward dance to excuse the existing missing definitions.. I really encourage you to make a ticket for this! There are ways around it, but I need 3 or 4 people endorsing this before moving to action..

cthoyt: Already done ;)

cthoyt: in my experience, the best thing to do is create a PR that enforces new standards, then allow slow contributions to that PR until the standards are met, then once everything has a definition, merge, then it's easy to enforce after

PR #445, opened 2021-03-19
characteristic-of part-of.. includious spruious sardine changes, Nico will fix

cmungall:
merged after #442.


cmungall (commented):
We all approve (RO annual call)

Issue #399, opened 2020-09-24
Adding new relationship "encircle"


teleaslamellatus:
So the exoskeleton (cuticle) in insects is composed of more and less flexible (sclerotized) regions (sclerites and conjunctivae), that are continuous with each other (they have fiat boundaries). Quite often, especially when the exoskeleton is evaginated, these structures "surround" each other. Like on the attached drawings, A "surrounds" B, B "surrounds" C, C "surrounds" D and D "surrounds" E.

This is basically the case for most insect appendages (e.g. antennae, legs, male genitalia).

I put "surround" in quotation marks because the RO terms surround and 'surrounded by', I think, should not be used in this case. It is like if you would like to say that the skin of the small finger is "surrounded by" the skin of the rest of your hand...perhaps. 

![IMG-1823 (1)](https://user-images.githubusercontent.com/9273125/94178073-98b0d100-fe68-11ea-90dd-8de971dc261e.JPG)
![IMG-1826 (1)](https://user-images.githubusercontent.com/9273125/94178076-98b0d100-fe68-11ea-9387-1efee27b02a4.JPG)
![IMG-1825 (1)](https://user-images.githubusercontent.com/9273125/94178078-99496780-fe68-11ea-902c-425adb04003f.JPG)
![IMG-1824 (1)](https://user-images.githubusercontent.com/9273125/94178079-99496780-fe68-11ea-85c9-3bfeb790066e.JPG)

ddooley:
Reminds me of varieties of enclave/exclave - and pene-exclave : https://en.wikipedia.org/wiki/Enclave_and_exclave

teleaslamellatus:
That sounds really interesting and perhaps reflects much better the situation than encircle.

matentzn:
Closing this as it seems @teleaslamellatus you have reflected on this issue.

matentzn:
Please reopen if I am wrong!

PR #456, opened 2021-04-21
added encircles and encircled by

JCGiron:
Added and annotated 'encircles' and 'encircled by' as requested at https://github.com/oborel/obo-relations/issues/399

### DIFF
--- a/src/ontology/ro-edit.owl
+++ b/src/ontology/ro-edit.owl
@@ -44,6 +44,8 @@ Declaration(Class(obo:RO_0002534))
 Declaration(Class(foaf:image))
 Declaration(ObjectProperty(obo:BFO_0000050))
 Declaration(ObjectProperty(obo:BFO_0000051))
+Declaration(ObjectProperty(obo:RO_0000093))
+Declaration(ObjectProperty(obo:RO_0000094))
 Declaration(ObjectProperty(obo:RO_0000300))
 Declaration(ObjectProperty(obo:RO_0000301))
 Declaration(ObjectProperty(obo:RO_0000302))
@@ -614,6 +616,7 @@ Declaration(AnnotationProperty(oboInOwl:created_by))
 Declaration(AnnotationProperty(oboInOwl:creation_date))
 Declaration(AnnotationProperty(oboInOwl:inSubset))
 Declaration(AnnotationProperty(rdfs:seeAlso))
+Declaration(AnnotationProperty(skos:closeMatch))
 ############################
 #   Annotation Properties
 ############################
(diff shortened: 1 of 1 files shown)


matentzn:
This pull request reveals a problem with something totally unrelated to your pull request @cmungall 
-> The annotations of the temporal relations seem to have been mangled to the same subset.

matentzn:
I am fixing that stay put.

src/ontology/ro-edit.owl
```diff
@@ -5127,6 +5148,7 @@ SubObjectPropertyOf(obo:RO_0002639 obo:RO_0002635)
 # Object Property: obo:RO_0002640 (intracellular endoparasite of)
 
 AnnotationAssertion(obo:IAO_0000115 obo:RO_0002640 "A sub-relation of endoparasite-of in which the parasite inhabits host cells.")
+AnnotationAssertion(obo:IAO_0000119 obo:RO_0002640 <https://en.wikipedia.org/wiki/Communal_roosting>)
```
matentzn, line 5151: Note that this annotation is not part of @JCGiron request, but an artefact from me reconciling the documentations. Please ignore during review - we want this annotation back independent of this pull request.

matentzn: Oh it was wrong, I removed it.

Clare72:
Can this be modified slightly so that z could be a 2D cross-section of 3D objects, x and y? I think your current definitions only cover 2D objects.

It would be nice to be able to say that the crepine encircles the mushroom body:
https://v2.virtualflybrain.org/org.geppetto.frontend/geppetto?id=VFB_00030606&i=VFB_00017894,VFB_00030867

JCGiron:
Yes!

How about: "x encircles y if both x and y are regions of the 2D structure z, or 2D cross-sections of a 3D object z, and the boundary of y on z is exclusively continuous with x."

@matentzn, should I edit that in the file and submit the pull request again? 

@Clare72 the 3D model in your link looks awesome!

matentzn:
Can you just edit this pull request here?

JCGiron:
I think I manage to edit this pull request (?)

matentzn:
Seems you did :)

balhoff:
@JCGiron we discussed this in the RO call, and wondered if you could discuss how these are different from surrounds/surrounded by. Would it make sense for these to be sub properties of surrounds/surrounded by? Could you add an rdfs:comment discussing the distinction?

teleaslamellatus:
"the shared boundary between x and y occupies the majority of the outermost boundary of x" part of the current definition of surrounded_by makes this class not feasible for encircled_by and I think because of this we came up with this new class (i.e. we were not able to use surrounded_by for our cases).

matentzn:
@teleaslamellatus thanks a ton.. These definitions really confuse me :D sorry to ask again, can you explain with an example exactly what the issue is? Is it the word "majority" that is the issue?

teleaslamellatus:
oops,

Sorry for not getting back earlier. So surrounded by is a subproperty of adjacent_to while encircles is a subproperty of continuous with. That was my major problems when proposed encircled by.

Clare72:
>Wouldn't it be sufficient to sub 'majority of' for 'all' to define encircled_by ?

Isn't that the opposite of what we want? I would expect 'encircles' to mean less coverage (e.g. ring on finger) than 'surrounds' (e.g. glove on hand).

dosumis:
> Isn't that the opposite of what we want? I would expect 'encircles' to mean less coverage (e.g. ring on finger) than 'surrounds' (e.g. glove on hand).

Good point - I guess it's asymmetric - all of the inner boundary of the ring is adjacent to my finger but only some of the outer boundary of my finger is adjacent to the inner boundary of the ring.  

In that case, encircled_by should be directly under  adjacent_to.

matentzn:
Please indicate action items here, thank you!

matentzn:
I converted this to draft - please change back to normal if ready for review and merging..

PR #463, opened 2021-05-26
New synapsing relationships

### LINKED ISSUE #455: Subproperties of 'synapsed to' and 'synapsed by'
Relationships to be used where there is a more meaningful number of synapses than just one.

Similar #379 for creation of RO:0013002 'receives synaptic input in' and RO:0013003 'sends synaptic output to'.

Ideal term labels would be 'receives synaptic input from' and 'sends synaptic output to', but 'sends synaptic output to' is already used for RO:0013003.

We could rename 'sends synaptic output to' -> 'sends synaptic output into' (which makes it clearer that it refers to a region)

Alternatively labels could be 'strongly synapsed to' and 'strongly synapsed by' or similar.

Definitions to mirror RO:0013002 'receives synaptic input in' and RO:0013003 'sends synaptic output to' as well as parent classes:
from relationship: "A relationship between an anatomical structure (including cells) and a neuron that has a functionally relevant number of chemical synapses to it."
to relationship: "A relationship between a neuron and an anatomical structure (including cells) that it has a functionally relevant number of chemical synapses to."

@dosumis

Clare72:
fixes #455 
After a little more thought, I went with 'synaptic input partner of' and 'synaptic output partner of' for labels, but open to other suggestions.


dosumis (changes requested):
All looks good apart from names, which I think are too confusing.  

Suggested fix:

Use names currently used for regions but modify to include object type in name

'receives synaptic input in'

\> 
 receives synaptic input from neuron
'receives synaptic input in region

'sends synaptic output to'
\> 
'sends synaptic output to neuron
'sends synaptic output to region

Clare72:
> 'sends synaptic output to neuron'

just thought of a problem with this - object of synapsed_to can be a muscle.

so maybe we just stick to adding 'region' to region relationships and use 'sends synaptic output to' and 'receives synaptic output from' for new relationships

matentzn:
What needs to be done here? We are scheduling an RO release for tomorrow, if you want this in, best finish :)

dosumis:
Maybe 'neuron' -> 'cell' ?

dosumis (approved):

Clare72:
So should we also change:
>between a neuron and an anatomical structure (including cells)

to:
>between a neuron and a cell

for 'sends synaptic output to cell'?

btw this would require terms for muscle cells (as we have in FBbt) and would be unsuitable for terms referring to 'complete' muscles.

dosumis:
> So should we also change:
> 
> between a neuron and an anatomical structure (including cells)
> 
> to:
> 
> between a neuron and a cell
> 
> for 'sends synaptic output to cell'?
> 
> btw this would require terms for muscle cells (as we have in FBbt) and would be unsuitable for terms referring to 'complete' muscles.

I think this is OK in the text def maybe with a gloss to clarify.  Strictly each individual NMJ is between a neuron and a (syncytial) muscle cell (myotube).  I don't know if a single motor neuron can form NMJs with multiple myotube.  I think we need to avoid having the range of the relation be 'cell' though, given we don't (can't) name individual cells in vertebrates.

matentzn:
Ready for merging?

dosumis (approved):

Issue #465, opened 2021-06-01
Provide structured form/robot template for new relationships


matentzn:
To avoid situations like #436 we should have a form to add new relationships, or maybe a robot template to make pull requests on. That way, we can:

1. reduce the strain on the community to deal with new term requests
2. Ensure that all new term requests are consistently labeled, have definitions and ORCIDS

As a first pass I suggest this:
1. We add an empty robot template somewhere with an example to fill in.
2. We add instructions how to fill in the template and make a pull request
3. We add a github action that checks for the changes to the template file, then
     1. compiling the file into OWL
     2. merging the OWL into ro-edit.owl
     3. on merge into master, delete the contents of the template again

matentzn:
This is now possible. For those that want to close this ticket, these are the instructions (please document in RO docs):

### Using ROBOT templates to add new relations

1. For each new new relation, add a row to https://github.com/oborel/obo-relations/blob/master/src/templates/new_relations.tsv
2. `TYPE` should be owl:ObjectProperty
3. Make a pull request as usual
4. GitHub actions will merge your changes into RO automatically

PR #466, opened 2021-06-01
DO NOT MERGE: new template based editing pipeline

matentzn:
cc @cthoyt @sierra-moxon 

see https://github.com/oborel/obo-relations/issues/465

Do not merge this, but review anyways. We should basically decide on minimal metadata, and implement the github action that does the deed.

Proposed (and implemented here) metadata:

|Field|Template|
|:---|:---|
|relation id (required)|ID|
|type|TYPE|
|label (required)|LABEL|
|definition (required)|A IAO:0000115|
|parent relation|SP %|
|creator ORCID (required)|AI oboInOwl:created_by|
|Date created (required)|AT oboInOwl:creation_date^^xsd:dateTime|
|relation characteristics|CHARACTERISTIC SPLIT=||
|domain|DOMAIN|
|range|RANGE|
|inverse property|IP %|
|example of use|A rdfs:comment|
|synonyms|A oboInOwl:hasExactSynonym SPLIT=||
|curator/usage notes|A IAO:0000116|

Can one of you fix up the github action (merge.yml) such that:

Trigger: Template tsv is changed

Action: 
1. the merge_template command is run (already in there)
2. The changes to ro-edit.owl are pushed to the branch

Thanks


cthoyt:
To push the files, you need something like this:

```yaml
      - name: Commit files
        run: |
          git add src/ontology/ro-edit.owl
          git config --local user.email "action@github.com"
          git config --local user.name "GitHub Action"
          git commit -m "Update ro-edit.owl"
      - name: Push changes
        uses: ad-m/github-push-action@master
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}
          branch: ${{ github.ref }}
```

.github/workflows/template.yml
```diff
@@ -0,0 +1,22 @@
...
+
+  workflow_dispatch:
+
+jobs:
+  merge_template:
+    runs-on: ubuntu-latest
+    container: obolibrary/odkfull:v1.2.26
+    steps:
+      - uses: actions/checkout@v2
+
```
cthoyt, line 17:
```suggestion
        with:
          persist-credentials: false # otherwise, the token used is the GITHUB_TOKEN, instead of your personal token
          fetch-depth: 0 # otherwise, you will failed to push refs to dest repo

````

File: .github/workflows/template.yml [hunk 2]
```diff
@@ -0,0 +1,22 @@
...
+      - name: Run ontology QC checks
+        env:
+          DEFAULT_BRANCH: master
+        run: cd src/ontology && make ROBOT_ENV='ROBOT_JAVA_ARGS=-Xmx6G' merge_template
+
```
cthoyt, line 22:
```suggestion
  - name: Commit files
        run: |
          git add src/ontology/ro-edit.owl
          git config --local user.email "action@github.com"
          git config --local user.name "GitHub Action"
          git commit -m "Update ro-edit.owl"
      - name: Push changes
        uses: ad-m/github-push-action@master
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}
          branch: ${{ github.ref }}
```

cmungall:
* example of use
 * synonyms
 * curator/usage notes

Some new relations will be fairly trivial compositions of existing relations and classes. We can have an annotation property to write down rules. I think SWRL makes most sense here. Defining relations is not OWL's strong point, and we shouldn't require people to understand obscure things like rolification.

matentzn:
@cthoyt wont that current action cause some kind of cycle?

It's triggered by a push, then changes something and pushes?

Is there any way we can restrict the triggers to only trigger when a specific file is changed?

Do I need to do something about that secret required for the push?

cthoyt:
@matentzn that's a good question. I'm pretty sure this doesn't cause an infinite trigger cycle because of some magic with the github actions push. If that's not the case, there's a way to add some pretty general conditional logic for when a given job shouldn't be run.

You don't need to do anything about the secret. It's always set to the correct thing by default by GitHub for you.

matentzn:
@cmungall I added your suggestions
@cthoyt last problem: do you know how I can, prior to running the merge_template command, reset ro-edit.owl to back to master? I hoped I could use git checkout directly in the action, but seems not..

cthoyt:
@matentzn a terrible hack would be to use curl or wget 🙈

matentzn:
I have spent too much time on this now, someone else will need to fix it :) Thanks for the help!

cthoyt:
Potential issue from 4 year old version of git installed in the docker image for obolibrary/odkfull:v1.2.26. Any chance that can get updated to use a modern version of git?

Screen Shot 2021-06-01 at 21 05 20

matentzn:
Super odd.. This is where this is defined:
https://github.com/INCATools/ontology-development-kit/blob/b85b2634a6bc8560f7659d58f0999640fb37e81e/Dockerfile#L16

I would have thought this will always install the latest version of git - despite the ubuntu base being comparatively old..

.github/workflows/template.yml
```diff
@@ -0,0 +1,44 @@
...
+    paths:
+      - 'src/templates/new_relations.tsv'
+  pull_request:
+    branches: [ master ]
+    paths:
+      - 'src/templates/new_relations.tsv'
+  workflow_dispatch:
+
+jobs:
+  reset_branch:
```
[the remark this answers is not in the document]
matentzn: In parallel? Oh man.. I thought at least they run in sequence..

matentzn:
I think we almost have it:

Somehow the token does not work:

```
- name: Push changes
        uses: ad-m/github-push-action@master
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}
          branch: ${{ github.ref }}
```

@balhoff @cthoyt can any idea?

cthoyt:
It looks like the operation of pulling in the old version of the ro-edit.owl messes up the branch tracking and lands you on `[detached HEAD 58f9432]`. It's unable to push because it's in a detached state

.github/workflows/template.yml
```diff
@@ -32,8 +32,4 @@ jobs:
           git config --local user.email "action@github.com"
           git config --local user.name "GitHub Action"
           git commit -m "Update ro-edit.owl"
-      - name: Push changes
-        uses: ad-m/github-push-action@master
-        with:
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          branch: ${{ github.ref }}
+          git push
```
cthoyt, line 35: *mic drops
*sounds airhorn
*scrambles the fighter jets

PR #490, opened 2021-08-27
add "has component participant" object property

### LINKED ISSUE #489: NR: add simple subproperty of has participant
In ENVO, a number of axioms use a cardinality restriction with [has participant](http://purl.obolibrary.org/obo/RO_0000057). But, this violated the DL profile b/c [has participant](http://purl.obolibrary.org/obo/RO_0000057)is a non-simple property. For background, see ENVO tickets [1072](https://github.com/EnvironmentOntology/envo/issues/1072) and [1182](https://github.com/EnvironmentOntology/envo/issues/1182).

One way to address this issue is the create a "fake" simple OP that is a subproperty of [has participant](http://purl.obolibrary.org/obo/RO_0000057) in a manner similar to the [has component](http://purl.obolibrary.org/obo/RO_0002180) OP.

Not sure what label should be. Some suggestions:
- has simple participant
- has direct participant
- has immediate participant

cc @cmungall @pbuttigieg @matentzn @balhoff

wdduncan:
Fixes #489 

This property is a "fake" OP that is used for cases in which `has participant` has a cardinality constraint. 

This ticket arose out of issues in ENVO. See ENVO tickets:  
https://github.com/EnvironmentOntology/envo/issues/1182   
https://github.com/EnvironmentOntology/envo/issues/1072

@cmungall I used your id range.

### DIFF
--- a/src/ontology/ro-edit.owl
+++ b/src/ontology/ro-edit.owl
@@ -150,6 +150,7 @@ Declaration(ObjectProperty(obo:RO_0002234))
 Declaration(ObjectProperty(obo:RO_0002235))
 Declaration(ObjectProperty(obo:RO_0002236))
 Declaration(ObjectProperty(obo:RO_0002237))
+Declaration(ObjectProperty(obo:RO_0002238))
 Declaration(ObjectProperty(obo:RO_0002240))
 Declaration(ObjectProperty(obo:RO_0002241))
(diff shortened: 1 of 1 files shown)


cmungall (approved):
Thanks!

We will need to make a page documenting all the component relations and place it here https://oborel.github.io/obo-relations/ but we can do later

PR #510, opened 2021-11-08
updated label name to 'disease arises from alteration in structure'

### LINKED ISSUE #508: update label of RO:0004030 to 'disease arises from alteration in structure'
The definition of RO:0004030 (disease arises from structure) is: 
A relationship between a disease and an anatomical structure where the material basis of the disease **is some pathological change in the structure.** Anatomical structure includes cellular and sub-cellular entities, such as chromosome and organelles.

I suggest updating the label to reflect that "the material basis of the disease is some pathological change in the structure."

Suggested label: disease arises from alteration in structure

sabrinatoro:
Fixes #508


nlharris (commented):
This looks ok to me, who needs to approve it?

matentzn (approved):

PR #511, opened 2021-11-09
Issue 497: add utilizes relation

### LINKED ISSUE #497: Relation for devices and the materials that the devices uses
Is there a relation that holds between a device, such as a hematology analyzer, and the specimen/sample that device uses to produce the result?  
Prov has a [used](https://www.w3.org/TR/prov-o/#used) predicate, but that holds between an activity and the thing the activity used.

Could add a `utilizes` relation? E.g, `hemotology-device#1` `utilizes` `blood-sample#1`

cc @cmungall @matentzn @balhoff @dosumis

wdduncan:
Fixes #497 

Added the `utilizes` relation, which is defined as:
```
X utilizes Y means X and Y are material entities, and X is capable of some process P that has input Y.
```
I also performed some other clean up on the `has component participant` relation that I add a bit ago.

@linikujp Can you please review the definition and usage example.


linikujp:
The definition is good for me.

PR #522, opened 2021-11-17
Add myristoylates relation

### LINKED ISSUE #521: Add "myristoylates" and helper property "is myristoyltransferase activity"
This issue is a simplified version of #266 that has a very specific focus.

As an addition to [RO:0002447 (phosphorylates)](http://purl.obolibrary.org/obo/RO_0002447) and [RO:0002480 (ubiquitinates)](http://purl.obolibrary.org/obo/RO_0002480), that occur under [RO:0002436 (molecularly interacts with)](http://purl.obolibrary.org/obo/RO_0002436), I would like to add a new relationship `myristoylates`. I've chosen this from the many possible molecular interaction relationships because it also has a corresponding GO molecular function and GO biological process, like `phosphorylates` and `ubiquitinates`.

Description:

> An interaction relation between x and y in which x catalyzes a reaction in which a myristoyl group is added to y.

Superproperty:

> 'capable of' o 'is myristoyltransferase activity' o 'has direct input'

Additional note:

The GO molecular function is imported into RO and this additional "subclass of" is added: `'is kinase activity' some Self` for phosphorylates. Therefore, the `myristoyltransferase activity` term will need to be imported from GO and appended with `'is myristoyltransferase activity' some Self`. Before closing this issue, it would be good to write here the correct way to import this GO term into RO.

## Comparison to previous molecular interaction relations

| Molecular Interaction Relation                                           | Helper Property                                                              | GO Molecular Function                                                                                      | GO Biological Process                                                                       |
|--------------------------------------------------------------------------|------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------|
| [RO:0002447 (phosphorylates)](http://purl.obolibrary.org/obo/RO_0002447) | [RO:0002481 (is kinase activity)](http://purl.obolibrary.org/obo/RO_0002481) | [GO:0016301 (kinase activity)](http://amigo.geneontology.org/amigo/term/GO:0016301)                        | [GO:0006468 (protein phosphorylation)](http://amigo.geneontology.org/amigo/term/GO:0006468) |
| [RO:0002480 (ubiquitinates)](http://purl.obolibrary.org/obo/RO_0002480)  | [RO:0002482 (is ubiquitination)](http://purl.obolibrary.org/obo/RO_0002482)  | [GO:0004842 (ubiquitin-protein transferase activity)](http://amigo.geneontology.org/amigo/term/GO:0004842) | [GO:0016567 (protein ubiquitination)](http://amigo.geneontology.org/amigo/term/GO:0016567)  |
| `myristoylates`                                                     | `is myristoyltransferase activity`                                   | [GO:0019107 (myristoyltransferase activity)](http://amigo.geneontology.org/amigo/term/GO:0019107)          | [GO:0018377 (protein myristoylation)](http://amigo.geneontology.org/amigo/term/GO:0018377)  |

The phosphorylates and ubiquitinates relations both have notes that they should later be axiomized to GO by @cmungall. Keeping track of these GO terms during curation will likely be helpful for the future when that happens.

cthoyt:
This PR closes #521 by adding a `myristoylates` relation as well as a `is myristoyltransferase activity` helper relation, following the pattern of [RO:0002447 (phosphorylates)](http://purl.obolibrary.org/obo/RO_0002447) and [RO:0002480 (ubiquitinates)](http://purl.obolibrary.org/obo/RO_0002480).

There are two things I didn't know how to do (help needed and would be appreciated!):

1. [x] How to choose local unique identifiers inside RO for these terms
2. [ ] How to import the related [GO:0019107 (myristoyltransferase activity)](http://amigo.geneontology.org/amigo/term/GO:0019107) from GO to inject `'is myristoyltransferase activity' some Self` to it

## See Also

- https://oborel.github.io/obo-relations/interaction-relations/

## Summary of discussion at the RO meeting (2022-01-11)

This is good in principle, but needs a template-based solution for both forwards and backwards versions of each relationship that can generate the OWL to reduce the need to do this manually.

Proof of concept for this is at https://github.com/cthoyt/obo-relations-interactions.

This also more highly motivates switching this repo over to ODK for dynamic imports of GO


cthoyt:
@matentzn mentioned that the way of importing GO terms isn't so great and it might be okay at the moment to leave this, so I will mark this PR as ready for review.

src/ontology/ro-edit.owl
```diff
@@ -6147,6 +6149,22 @@ AnnotationAssertion(rdfs:label obo:RO_0016002 "has disease")
 AnnotationAssertion(rdfs:seeAlso obo:RO_0016002 "https://github.com/oborel/obo-relations/issues/478")
 SubObjectPropertyOf(obo:RO_0016002 obo:RO_0016001)
 
+# Object Property: obo:RO_0018001 (is myristoyltransferase activity)
+
+AnnotationAssertion(obo:IAO_0000117 obo:RO_0018001 "Charles Tapley Hoyt"@en)
+AnnotationAssertion(dce:contributor obo:RO_0018001 <https://orcid.org/0000-0003-4423-4370>)
+AnnotationAssertion(rdfs:label obo:RO_0018001 "is myristoyltransferase activity"@en)
+SubObjectPropertyOf(obo:RO_0018001 obo:RO_0002564)
```
matentzn, line 6157: use rdfs:seeAlso to link the related github issue

src/ontology/ro-edit.owl
```diff
@@ -6147,6 +6149,22 @@ AnnotationAssertion(rdfs:label obo:RO_0016002 "has disease")
 AnnotationAssertion(rdfs:seeAlso obo:RO_0016002 "https://github.com/oborel/obo-relations/issues/478")
 SubObjectPropertyOf(obo:RO_0016002 obo:RO_0016001)
 
+# Object Property: obo:RO_0018001 (is myristoyltransferase activity)
```
matentzn, line 6152: Is this reflexive? we would want to see something like domains and ranges here. Usually, relations starting with `is` and not ending with `of` are reflexive, i.e. relating an individual to itself. `is person`.

src/ontology/ro-edit.owl
```diff
@@ -6147,6 +6149,22 @@ AnnotationAssertion(rdfs:label obo:RO_0016002 "has disease")
...
+
+AnnotationAssertion(obo:IAO_0000117 obo:RO_0018001 "Charles Tapley Hoyt"@en)
+AnnotationAssertion(dce:contributor obo:RO_0018001 <https://orcid.org/0000-0003-4423-4370>)
+AnnotationAssertion(rdfs:label obo:RO_0018001 "is myristoyltransferase activity"@en)
+SubObjectPropertyOf(obo:RO_0018001 obo:RO_0002564)
+
+# Object Property: obo:RO_0018002 (myristoylates)
+
+AnnotationAssertion(obo:IAO_0000116 obo:RO_0018002 "Axiomatization to GO to be added later"@en)
```
matentzn, line 6161: Not sure how useful this is without a reference to a GitHub issue.

matentzn:
usually we do id ranges in a separate pr so no one snatches yours... this might have even happened, since there is a conflict..

matentzn:
@balhoff can you give this a read while Chris is MIA?

src/ontology/ro-edit.owl
cthoyt: This was following the same form of the phosphorylates and ubiqutinates, so this question is also valid for those. I think we need to pull in Chris to reconsider some of these things that got bumped for later

src/ontology/ro-edit.owl
cthoyt: Added reference to github issue in ae78197, but this note is the same as the one @cmungall left in the phosphorylates/ubiquitinates relations too

src/ontology/ro-edit.owl
cthoyt: Added in ae78197

src/ontology/ro-edit.owl
matentzn: Good to learn from the generals :D But we can do some things a bit better I believe.

src/ontology/ro-edit.owl
dosumis: If I understand correctly, the helper relation is only there for use in a property chain that defines **myristoylates**.  It sits under:

rdfs:label "helper property (not for use in curation)"
'curator note' "This property or its subproperties is not to be used directly. These properties exist as helper properties that are used to support OWL reasoning."

I think the logic makes sense and the precedent with phosphorylates and ubiquitinates supports adding it.

src/ontology/ro-edit.owl
dosumis: Ideally we would have dynamic import of GO driving this.  @matentzn  - how far are we from having this in place?

dosumis:
Looks good to me, assuming no ID clashes.

cthoyt:
So the follow-up to this PR will be another one that drops on the order of tens of similar relationships for other PTM relationships, so I think it would be nice to address some of the questions here before merging this one. I'm okay to wait until the RO meeting in early january where we can hammer this out if nobody has any ideas on where to start now

cmungall:
I am happy with adding these incrementally. We need a better mechanism for bringing in these GO class-as-predicate shortcuts but I don't want to hold this PR up on that.

cmungall:
superseded by #586

PR #525, opened 2021-11-18
Weakening range of causal relation between processes. This may be temporary. See https://github.com/oborel/obo-relations/issues/524

### LINKED ISSUE #524: relax range constraint on regulates
change from process to process or characteristic

(there are other tickets to be linked here but I can't find them)

This change is necessary for go-plus to be coherent (currently crytpic incoherence as not caught by elk 0.43)

alternative:

create three parallel relations and use these in GO for the RoBQ branch

The two alternatives are not mutually exclusive - I think it is most practical to first relax, then we can make the GO change. But this is actually quite urgent

cmungall:
Fixes #524

### DIFF
--- a/src/ontology/ro-edit.owl
+++ b/src/ontology/ro-edit.owl
@@ -4272,7 +4272,7 @@ AnnotationAssertion(obo:IAO_0000232 obo:RO_0002501 "Do not use this relation dir
 AnnotationAssertion(rdfs:label obo:RO_0002501 "causal relation between processes")
 SubObjectPropertyOf(obo:RO_0002501 obo:RO_0002410)
 ObjectPropertyDomain(obo:RO_0002501 obo:BFO_0000003)
-ObjectPropertyRange(obo:RO_0002501 obo:BFO_0000003)
+ObjectPropertyRange(Annotation(oboInOwl:editor_note "The range on this relation is weaker than the name suggests, and it can be used to connect processes to qualities, where those qualities are proxies for processes. See https://github.com/oborel/obo-relations/issues/524 for a discussion.") obo:RO_0002501 ObjectUnionOf(obo:BFO_0000003 obo:BFO_0000019))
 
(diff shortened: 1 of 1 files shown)


matentzn (approved):

src/ontology/ro-edit.owl
```diff
@@ -4271,8 +4271,8 @@ AnnotationAssertion(obo:IAO_0000119 obo:RO_0002501 <http://purl.obolibrary.org/o
 AnnotationAssertion(obo:IAO_0000232 obo:RO_0002501 "Do not use this relation directly. It is intended as a grouping for a diverse set of relations, all involving cause and effect.")
 AnnotationAssertion(rdfs:label obo:RO_0002501 "causal relation between processes")
 SubObjectPropertyOf(obo:RO_0002501 obo:RO_0002410)
-ObjectPropertyDomain(obo:RO_0002501 obo:BFO_0000003)
-ObjectPropertyRange(obo:RO_0002501 obo:BFO_0000003)
+ObjectPropertyDomain(obo:RO_0002501 ObjectUnionOf(obo:BFO_0000015 obo:BFO_0000019))
```
matentzn, line 4274: Its odd you go from BFO:3 to BFO:15 or BFO:19. Is that intended?

matentzn (changes requested):
Just double check my comment

balhoff (changes requested):
Just same question as Nico, otherwise looks good.

cmungall:
occurent => process or quality.

yes this is intentional. Are there any useful siblings of process?

balhoff (approved):

balhoff (changes requested):
If we're going to do this, we will need to do the same to 'precedes' and 'temporally related to', and 'preceded by', etc.

balhoff:
Please hold off on merging this; I am going to make an alternative PR.

Issue #467, opened 2021-06-18
bridge CHEBI:role to BFO:role


balhoff:
In GO there is an axiom: [role](http://purl.obolibrary.org/obo/BFO_0000023) EquivalentTo [role](http://purl.obolibrary.org/obo/CHEBI_50906)

This equates the role class in CHEBI to the one used in RO, and catches cases when chemical roles (specifically dependent continuants) are accidentally used as chemicals (material entities).

When I merge several OBO ontologies together, like GO, Uberon, MP, and HP, I often run into disjointness violations with use of chemical roles. Generally this is because this bridging axiom is randomly placed in GO and not imported into other ontologies.

Can we add this axiom to RO? It would be nice to have this shared across all these ontologies. Maybe this will be the sort of thing COB is used for down the road, but this is an immediate problem and RO is a nice central place.

I'm putting an example explanation for one of these problems in the next comment.

balhoff:
## [aminoaciduria](http://purl.obolibrary.org/obo/MP_0003311) EquivalentTo [Nothing](http://www.w3.org/2002/07/owl#Nothing) ##

  - [aminoaciduria](http://purl.obolibrary.org/obo/MP_0003311) SubClassOf [aciduria](http://purl.obolibrary.org/obo/MP_0010028)
    - [aciduria](http://purl.obolibrary.org/obo/MP_0010028) EquivalentTo [has part](http://purl.obolibrary.org/obo/BFO_0000051) some 
([increased amount](http://purl.obolibrary.org/obo/PATO_0000470) and ([inheres in](http://purl.obolibrary.org/obo/RO_0000052) some 
([acid](http://purl.obolibrary.org/obo/CHEBI_37527) and ([part of](http://purl.obolibrary.org/obo/BFO_0000050) some [urine](http://purl.obolibrary.org/obo/UBERON_0001088)))) and ([has modifier](http://purl.obolibrary.org/obo/RO_0002573) some [abnormal](http://purl.obolibrary.org/obo/PATO_0000460)))
      - [acid](http://purl.obolibrary.org/obo/CHEBI_37527) SubClassOf [chemical role](http://purl.obolibrary.org/obo/CHEBI_51086)
        - [chemical role](http://purl.obolibrary.org/obo/CHEBI_51086) SubClassOf [role](http://purl.obolibrary.org/obo/CHEBI_50906)
[...]

PR #528, opened 2021-11-23
Bridge BFO role to CHEBI role.

balhoff:
Addresses issue #467. This is also taken care of by the range declaration for [has role](http://purl.obolibrary.org/obo/RO_0000087), however we are still not applying range reasoning in very many places due to most people still using ELK 0.4.3.


matentzn:
Is RO the right place for this kind of alignment? I would imagine that once we move to ODK, much of this kind of axiomatisation should move to COB, right? In any case, I think its the right move, but this is probably one for @cmungall to judge.

matentzn:
Also: since our CI pipelines use Elk 0.4.3, this wont really solve our "role part of material entity" problem, right?

balhoff:
> Also: since our CI pipelines use Elk 0.4.3, this wont really solve our "role part of material entity" problem, right?

This is meant to solve that, at least in part. `has role` has a range axiom that would make this less needed if the reasoner supported ranges.

balhoff:
Changed to be subclass of PATO:quality, which is what I think is meant by "characteristic" above.

cmungall:
correct https://github.com/OBOFoundry/COB/blob/master/src/ontology/components/cob-to-external.tsv#L120

Issue #477, opened 2021-08-03
NTR: Has exposure medium 


laurenechan:
@diatomsRcool @cmungall 

Hello,
Within the Environmental Conditions, Treatments, and Exposures Ontology (ECTO) we are using the higher level model of Exposure Ontology (ExO) to develop our exposure event classes. This includes documenting the route, stressor, and medium of the exposure when available. 
We are currently using relations including 'has exposure stressor' and 'has exposure receptor', but we do not have a relation for 'has exposure medium'. 

We propose the following details for this term:
Label: has exposure medium
Parent: related via exposure (RO:0002244)
Definition: A broad relationship between an exposure event or process and any entity (e.g., an organism or environmental substance) through which something else is transmitted or carried during an exposure event.
Definition source: ExO_0000083

laurenechan:
I'm wondering if the 'occurs in' object property would give the impression that the exposure is occurring in a specific location or medium. 
Exposure medium is defined in ExO as 'a substance through which something else is transmitted or carried during an exposure.', with some examples being soil, water, air, and food. Exposure mediums provide information about what entity is containing the exposure stimulus. I don't think this means the medium encompasses the location of the actual exposure itself whether that be a physical location or even that the exposure is happening in the medium being documented.

mellybelly:
Could be a subproperty?

laurenechan:
The exposure medium is not intended to pick out the location where the exposure occurs (e.g. lungs vs air vs room), but rather denote the entity that is allowing for the stimulus to be transmitted to the actual site of the exposure. So in your example, the medium would be air, stimulus is the smoke, and the location of the exposure is the lungs (because you and your lungs are the exposure receptor).

laurenechan:
Hmm we currently use the second approach offered, but I wonder if it would be more meaningful to assign the medium to the stimulus itself instead of the exposure event generally? thoughts @diatomsRcool ?

Also spitballing on the first proposal, 
If we had the stimulus and medium packaged up together so to say, and then you had a second stimulus and the same medium packaged up, that might allow us to showcase simultaneous exposures within one exposure event without a mixture. Would that be of interest to us? For example if you have general pollution in the air that you breathe in, but you also have welding fumes at your work that you breathe. Two mixtures that are not technically one super big mixture, could we represent them as simultaneous exposures to separate stimuli within the same medium? 
Would also be interesting for nutrient exposures in food.

wdduncan:
You can also have relations for both situations:
[...]

PR #534, opened 2021-12-14
Adding "has exposure medium" relation

laurenechan:
Adding the relation 'has exposure medium' as discussed in issue #477


cmungall (approved):

src/ontology/ro-edit.owl
```diff
@@ -6147,6 +6148,15 @@ AnnotationAssertion(rdfs:label obo:RO_0016002 "has disease")
 AnnotationAssertion(rdfs:seeAlso obo:RO_0016002 "https://github.com/oborel/obo-relations/issues/478")
 SubObjectPropertyOf(obo:RO_0016002 obo:RO_0016001)
 
+# Object Property: obo:RO_0016004 (has exposure medium)
+
+AnnotationAssertion(obo:IAO_0000115 obo:RO_0016004 "X has exposure medium Y if X is an exposure event (process), Y is a material entity, and the stimulus for X is transmitted or carried in Y.")
+AnnotationAssertion(obo:IAO_0000119 obo:RO_0016004 "ExO_0000083")
```
cmungall, line 6154: This is just a string, and not even a CURIE

I would recommend instead annotating the definition axiom - or if the current annotation datamodel is used then use a URL instead

src/ontology/ro-edit.owl
```diff
@@ -6147,6 +6148,15 @@ AnnotationAssertion(rdfs:label obo:RO_0016002 "has disease")
...
+AnnotationAssertion(dce:date obo:RO_0016004 "2021-12-14T20:41:45Z"^^xsd:dateTime)
+AnnotationAssertion(oboInOwl:created_by obo:RO_0016004 <http://orcid.org/0000-0002-7463-6306>)
```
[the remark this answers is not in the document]
laurenechan: Adjusted the orcid link to https

src/ontology/ro-edit.owl
laurenechan: I moved the 'definition source' annotation onto the definition itself, hopefully that is the change you were intending to see. Please let me know if not!

PR #537, opened 2021-12-15
Fixes for branching structures

cmungall:
Fixes error noted by @dosumis on CL/hubmap call This fix brings us into alignment with https://github.com/obophenotype/uberon/wiki/Branching-structures-Design-Pattern

See also https://github.com/obophenotype/uberon/issues/2212


matentzn:
It would be sort of not bad if we had some kind of test that in RO that tests changes on key OBO ontologies..

dosumis:
> It would be sort of not bad if we had some kind of test that in RO that tests changes on key OBO ontologies..

+💯 !  It's worrying If this came in through recent maintenance.  Lots of capacity to screw up many downstream ontologies.

dosumis (approved):

matentzn:
QC says NOPE:

image

cmungall:
sometimes I really hate owl...

OK, this PR introduces an annotation property:

http://purl.obolibrary.org/obo/RO_0004036

image

Using this instead

the DL restriction is really dumb, as having these as logical axioms is really useful and completely tractable for RL reasoning

PR #540, opened 2021-12-15
Add prefix `orcid` and migrate previous usage of ORCiD identifiers

cthoyt:
Following the current UBERON sprint to make the contribution metadata more useful, this PR does the following:

1. Adds the `orcid` prefix
2. Normalizes all ORCID usages to HTTPS (some were HTTP before)
3. Switch to CURIE forms instead of IRI forms
4. Does not update the way that @wdduncan annotated his entries. This happens twice and looks like this:
   ```
   AnnotationAssertion(obo:IAO_0000117 obo:RO_0002238 "Bill Duncan (https://orcid.org/0000-0001-9625-1899)")
   AnnotationAssertion(obo:IAO_0000117 obo:RO_0017001 "Bill Duncan (https://orcid.org/0000-0001-9625-1899)")
   ```

   I think this PR might be a good place to continue the discussion on what the best practices in this ontology might be going forwards.


matentzn:
Did you try to open in Protege and save to see what happens?

cthoyt:
Do you expect something bizarre to happen? It seems to open and save alright for me.

wdduncan:
@cthoyt Yes. We should continue this discussion. I've raised this point before. I am not in favor of only having an ORCID. When I am interested in who created/edited a term, I am interested in the person's name (often I know them), and looking up the name is a hassle.  
Also, not everyone has an ORCID.

cthoyt:
@wdduncan I agree, there might a compromise where we save both the ORCID for programmatic consumption and the label for human consumption.

wrt ORCiDs: I think it's totally fair to impose a rule that all contributors should have one (going forwards). It's free and only takes 2 minutes to make one - and one don't actually have to make any information besides one's name publicly available if that's undesirable.

matentzn:
@shawntanzk can you check this? Checkout the branch and look what happens if you make a change in protege and save it? If its fine, I think we should make this the standard across all our ontologies.

wdduncan:
@cthoyt I'm not opposed to the rule of requiring an ORCID. This would be a good topic to bring up on the RO call. You never know ... someone may have an issue with such a rule.

Also, were there terms that pre-dated ORCID? If so, were those updated too?

matentzn:
We are doing this massive pushing migrating thousands of annotations to this format for Uberon:
https://github.com/obophenotype/uberon/pull/2206

shawntanzk:
> @shawntanzk can you check this? Checkout the branch and look what happens if you make a change in protege and save it? If its fine, I think we should make this the standard across all our ontologies.

When I save it on protege, it reverts the changes back
Screenshot 2021-12-15 at 20 07 51

balhoff:
It might be due to protection against illegal CURIES in RDF-XML, which I believe cannot begin with a digit in the local part (they may be used as XML element names).

matentzn:
There we go, I knew there was some problem :P thanks for checking and illuminating @shawntanzk and @balhoff

cthoyt:
Would there still be a benefit to adding the prefix definition?

matentzn:
I think it is just redundant - what we want is a SHEX/SHACL validation process for ORCIDs.

cthoyt:
> 

Okay, then maybe the only part of this PR that's useful is normalizing the HTTP to HTTPS, but we should close this one now in favor of a different PR that does what you suggested.

matentzn:
Dont worry for the https issue we have a different process that will blaze through all ontologies

PR #543, opened 2021-12-17
Add 'regulates characteristic' relations.

balhoff:
Adding three relations between processes and characteristics/qualities they regulate or result in. These are needed for the 'regulation of biological quality' branch of GO, in order to fix incorrect usage of 'regulates' there (see https://github.com/geneontology/go-ontology/issues/16524):

- 'regulates characteristic'
- 'regulates increase in characteristic'
- 'regulates decrease in characteristic'

This PR would supersede #525.


balhoff:
@ukemi suggests changing the parent property label to 'regulates manifestation of characteristic', which I agree with.

cmungall (approved):

dosumis:
prefer "positively regulates characteristic" , "regulate increase in" sounds odd when you substitute a specific charcteristic e.g. "regulates increase in membrane potential"  vs "positively regulates membrane potential"

PR #545, opened 2022-01-10
Added properties for marker set

shawntanzk:
For Brain Data Standards (and probably future transcriptomic cell typing), we need a property that describes a set of markers that is able to identify the cell type based on certain algorithms. These are usually context-dependent and hence we also need a property to link the context (in our case anatomical) - we therefore are suggesting 'has characterizing marker set' and 'has anatomical context'

Personally don't like the use of the word "properties" in has_anatomical_context, but can't think of a better word - very open to suggestions.


shawntanzk:
@patrick-lloyd-ray & @tgbugs - not sure if I can tag you here but figured you might be interested in this

cmungall (commented):
Domain and Range constraints?

cthoyt (commented):
Would be nice if the PR had a better description for anyone reading through (or in the future)

addiehl:
A set of transcribed genes or markers might be considered to be a 'signature' for a particular cell type or set of related cell types.

addiehl:
Thus, 'has marker signature set' (or 'has marker signature') might be a better name or alternative name for the proposed relation 'has characterizing marker set'

shawntanzk:
> Domain and Range constraints?

has_characterizing_marker_set - Domain: cell (or material entity?), range:? <- might need some help here @dosumis, we will be using an individual that hold genes but im not 100% sure what I should put that under? 

has_anatomical_context - Domain: ? same as above with range, range: 'anatomical entity'

addiehl:
Cells can have multiple signatures, that may reflect combinations of transcriptional programs.  We see this all the time in the immune system, with for instance both T cell subsets and ILC subsets express Th1, Th2, Th17 etc. cytokine patterns, overlaid with the presence or absence of transcriptional programs for cytoxicity (expression of perforin, granzymes), and sometimes 'regulatory'-phenotype subsets (IL25, FoxP3).

shawntanzk:
Drop anatomical context for now, leave range for characterizing_marker_set empty for now -> we need a term for this somewhere

wdduncan:
@addiehl Are you proposing that there needs to be `cell signature` class? If so, then sounds like a collect/aggregate with the proteins as members. If not, then I suppose the `has-cell-signature` relation would hold between the cell and the various proteins.

shawntanzk:
@addiehl we have decided to keep the label as is, but have added "has marker signature set" as a synonym

@wdduncan I think that's my bad for not writing this ticket up, but basically, the model we are aiming for is as follows: 

Cell
subClassOf: (has_characterizing_marker_set some S1)  

S1 has_part some gene 1
S1 has_part some gene 2
S1 has_part some gene 3
S1 has_context some M1

wdduncan:
Thanks @shawntanzk 

> Cell
> subClassOf: (has_characterizing_marker_set some S1)
> 
> S1 has_part some gene 1
> S1 has_part some gene 2
> S1 has_part some gene 3
> S1 has_context some M1

What is the type of `S1`?

addiehl:
Over the past decade, I've periodically mentioned the idea of having a whole ontology of cell signature classes. I guess it will effectively be built with this work, though under a somewhat inferior name, IMO.

shawntanzk:
@wdduncan - currently we are still deciding on that - we have a call soon with the brain data standards ontology group to discuss that (ps any recommendation would be amazing), it is also why the range is empty at the moment. I think we would ideally like to use a term in an upper ontology or something to group them, but we can't find anything suitable. Our temp fix is to use a grouping term "Gene_set_class"

> though under a somewhat inferior name, IMO.

LOL! names are always difficult >.<

addiehl:
To make this more useful, the pattern needs to allow proteins in addition to genes to be applicable to data beyond scRNA seq.

Cell
subClassOf: (has_characterizing_marker_set some S2)

S2 has_part some protein 1
S2 has_part some protein 2
S2 has_part some protein 3
S2 has_context some M1

In addition to figuring out what S1 and S2 are, what is M1, the context?

Importantly, when we reference gene 1, gene 2, etc., we are using shorthand to refer to the transcripts of gene 1, gene 2, etc, which is imperfect from an ontological perspective.

Terms for 'gene expression profile' and 'protein expression profile' exist in NCIT, EDAM, and FBcv.  A profile seems like an information artifact, perhaps a 'data set having parts 'measurement datum'(s) about the expression of particular genes or proteins.

shawntanzk:
@addiehl I think this pattern does allow for proteins too - will make sure that the range allows for this. I think our aim is for the range to be more generic rather than gene or protein specific. But yeah basically figuring out what S1 and S2 are categorized under is what we need to figure out I guess.

M1 in our case would be anatomical context (eg gene set S1 characterizes the cell type specifically in the motor cortex for example) but could also be extended to things like species or whatever is required (I'd imagine immune cells might need context such as activation state or something?)

addiehl:
A disease state might be a context, as well as an organismal state, such as sleeping or being awake. So context is a pretty broad category.

patrick-lloyd-ray:
Apologies for jumping in late here, @shawntanzk. Also apologies for any typos/simple mistakes here as I'm not thinking very clearly at the moment. 

@addiehl is correct that 'context' is very broad. Formal semanticists typically use the term 'context' to mean something like _situation of utterance_ or the way things are that helps determine what is expressed by an utterance. I assume by using 'anatomical context' you/we intend to constrain the 'things' in the previous sentence to anatomical entities, right? That's how we get the range for `has_anatomical_context` -- otherwise the range will be `thing` which is as broad as it gets. If that's correct, then maybe we need `has_context` and `has_anatomical_context` as well. This would allow others to follow the pattern to specify different contexts (as needed) in the future (has_[X]_context), where domain will be information artifact and range will be determined by [X]. 

As far as S1/S2, I think the suggestion that could they be object aggregates is good, if we are talking about the actual proteins (@wdduncan and @addiehl 's suggestion). However, object aggregates wouldn't have contexts (in the above sense), so I'm not convinced that's what is needed here. If 'context' has some other meaning, then this objection is moot.

If we are meaning to refer to information artifacts, then it's more complicated (as @addiehl points out). But I don't know why, off the top of my head, why that couldn't be `data item` or `information artifact`. In fact, I think `data item` (or child) will be what works if we also want it to be the domain of the `has_anatomical_context` relation as well. We'd have to change the pattern a bit, of course, if that were the case. Thoughts?

shawntanzk:
@patrick-lloyd-ray yeah, I think `has_anatomical_context` is sort of a constraint, though because of how we couple it to cells which have anatomical axioms on them, we aren't planning on using `has_anatomical_context` for now. I do agree that having a parent term `has_context` with  `has_anatomical_context` under it would be good for future expansion of this term. I'll open a ticket for that (we have removed `has_anatomical_context` to not bog down this PR since we aren't using it for now). 

I think we did look at object aggregates, but weren't quite happy with it. probably needs more discussion. Not against `data item`/what @addiehl suggested in the quoted text below, but that is a bit beyond my knowledge at this point >.< not 100% sure I understand what that means and all. 

>  A profile seems like an information artifact, perhaps a 'data set having parts 'measurement datum'(s) about the expression of particular genes or proteins.

wdduncan:
I understand what 'context' means from a linguistic perspective. I'm not sure how you using it an ontological sense. Can you give a concrete example of how you wanting to use `has_anatomical_context` (or something similar) and how the use it modifies the semantics (or our understanding ) of the cell's signature?

src/ontology/ro-edit.owl
```diff
@@ -6147,6 +6149,17 @@ AnnotationAssertion(rdfs:label obo:RO_0015003 "subcluster of")
 SubObjectPropertyOf(obo:RO_0015003 obo:BFO_0000050)
 TransitiveObjectProperty(obo:RO_0015003)
 
+# Object Property: obo:RO_0015004 (has characterizing marker set)
+
+AnnotationAssertion(obo:IAO_0000112 obo:RO_0015004 "transcriptomically defined cell type X subClass_of: (has_characterizing_marker_set some S1); S1 has_part some gene 1, S1 has_part some gene 2, S1 has_part some gene 3, S1 has_anatomical_context some [region]")
+AnnotationAssertion(obo:IAO_0000115 obo:RO_0015004 "A relation between a material entity and a set of markers that indentifies it in a particular context.")
```
dosumis, line 6155: Suggested def:

A relation that applies between a cell type and a set of markers that can be used to uniquely identify that cell type.

src/ontology/ro-edit.owl
```diff
@@ -6147,6 +6149,17 @@ AnnotationAssertion(rdfs:label obo:RO_0015003 "subcluster of")
...
+AnnotationAssertion(obo:IAO_0000116 obo:RO_0015004 "C has_characterizing_marker_set y iff x is a context in which C resides and y is a set of markers that is sufficient to identify C in context x.")
+AnnotationAssertion(dce:contributor obo:RO_0015004 <https://orcid.org/0000-0001-7258-9596>)
```
dosumis, line 6157: Not sure we need something so formal, but if we keep this - suggest relaxing to

C has_characterizing_marker_set y iff: C is a cell type and y is a collection of genes or proteins whose expression is sufficient to distinguish cell type C from most or all other cell types.

dosumis: Also odd that the more formal bit is in editor's note.

It would also be good to have some guidance e.g. do not use this for cases where set of genes/proteins are only useful as markers in some specific context - e.g. in some specific location.  In these cases it is recommended to make a more specific cell class restricted to the relevant context.

dosumis (changes requested):
I suggest removing the ability to specify context.

See comments in-line for my suggestions on how to achieve this.

shawntanzk:
Thanks everyone for the discussion - it really helped in shaping this better. We have now changed it to require implicit constraint instead of having has_anatomical_context, this should be easier to model. I've also followed @dosumis suggestions on the changes required for this in the PR, please do have a look and if you're good with that, please provide a thumbs up in the PR review. Thanks!

wdduncan:
@shawntanzk in the `example of usge`:
```
transcriptomically defined cell type X subClass_of: (has_characterizing_marker_set some S1); S1 has_part some gene 1, S1 has_part some gene 2, S1 has_part some gene 3.
```
Could you replace the variables `X` and `S1` with an actual real world example? You can also add another `example of usage` annotation that has a concrete example.

shawntanzk:
@wdduncan thanks for the suggestion, I've added it to the PR (It's, unfortunately, a bit complicated looking, so I think it's a good idea to have both)

src/ontology/ro-edit.owl
```diff
@@ -541,6 +541,8 @@ Declaration(ObjectProperty(obo:RO_0014001))
 Declaration(ObjectProperty(obo:RO_0015001))
 Declaration(ObjectProperty(obo:RO_0015002))
 Declaration(ObjectProperty(obo:RO_0015003))
+Declaration(ObjectProperty(obo:RO_0015004))
+Declaration(ObjectProperty(obo:RO_0015005))
```
shawntanzk, line 545: Note to self - this declaration needs to be removed I think, will check

shawntanzk:
Plan is to add range as SO:0001260 (sequence collection) - this encapsulates both proteins and genes (not sure how to handle imports in RO so waiting for answer on that.

shawntanzk:
Range will be added later after RO is moved to ODK - imports are tricky without ODK-ification 
PR all ready now, I will wait for at least one positive review before merging this in

dosumis (approved):

dosumis:
If no further objections we will merge this PR on 4th Feb.

PR #551, opened 2022-01-14
DO NOT MERGE: New AP for unilateral injection

### LINKED ISSUE #550: Annotation property for tagging unilaterally injected axioms
See https://github.com/OBOFoundry/OBOFoundry.github.io/issues/1443#issuecomment-1012942543

label: modifies_externally_defined_OWL_entity
def: This axiom alters the meaning of an OWL Entity, but has not been officially sanctioned by the editors of the ontology to which this entity belongs.
range: Boolean

dosumis:
Fixes #550

Have added OWL axiom as domain.  Think it would be good if this is can restrict usage, but don't know if that might have unintended consequences (e.g. due to possibility of alternative reification patterns using RDF standards)

### DIFF
--- a/src/ontology/ro-edit.owl
+++ b/src/ontology/ro-edit.owl
@@ -650,10 +650,28 @@ SubAnnotationPropertyOf(obo:RO_0001900 obo:RO_0002422)
 
 # Annotation Property: obo:RO_0002027 (modifies externally defined OWL Entity)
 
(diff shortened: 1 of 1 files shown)


cthoyt:
@dosumis can you include a description on how someone would go about using this?

dosumis:
Have added attempt at justification & an example.

PR #556, opened 2022-01-19
Release 2022-01-19.

balhoff:
@matentzn @shawntanzk did I do this right? Just did `make release` and committed. Then made a tag.


cmungall (approved):
Looks as expected!

Issue #526, opened 2021-11-21
Documentation request: how to join the RO meetings


matentzn:
For current and prospective RO users it would be useful to know how to learn about and join the quarterly RO meetings. This should be added to the documentation https://oborel.github.io/obo-relations/

nlharris:
I think the answer is that you should request to join the obo-relations mailing list (which is mentioned on https://github.com/oborel/obo-relations, but that README doesn't mention the quarterly meetings -- should it?).

matentzn:
Yes it should. We should:

- [ ] Update https://github.com/oborel/obo-relations/blob/master/CONTRIBUTING.md to include a mention to the regular meetings and how to learn about them, and add a bullet on how to join the RO mailing list
- [ ] We need to be sure that RO mailing list is actually used to announce meetings in advance.. The person dedicated to organising the event should take care of that, not sure how to make sure they actually do.
- [ ] We need to add a link from the main README to https://github.com/oborel/obo-relations/blob/master/CONTRIBUTING.md "How to contribute to RO and join the RO working group".

nlharris:
Thanks for the specifics! I have been announcing RO meetings on the mailing list. The other two tasks I just made PRs for.

matentzn:
This is addressed now, thanks for all that @nlharris

PR #558, opened 2022-01-19
added "How to contribute to RO and join the RO working group"

nlharris:
Addresses part of #526


matentzn (approved):
Great, thanks @nlharris

PR #567, opened 2022-02-01
Moving PATO OP to RO

shawntanzk:
see https://github.com/pato-ontology/pato/issues/454
Note: `has relative magnitude` does not have a definition in PATO so might need to write one up here


matentzn:
Did you change any of the PATO axiomatisation? I mean for this PR here, did you add anything?

shawntanzk:
@matentzn  nope - should be exactly the same as how it is in pato

shawntanzk:
oh except i changed the xref to orcid

balhoff (approved):

PR #568, opened 2022-02-07
2022-02-07 Release

shawntanzk:
Need a release for some new terms I added. 
@matentzn I know you don't review content, but adding you as a reviewer cause RO isn't ODK and want to make sure it's all good


matentzn (approved):
"has relative magnitude" really needs a definition and/or a comment, but no blocker here.

shawntanzk:
> "has relative magnitude" really needs a definition and/or a comment, but no blocker here.

I'll make a ticket for it in RO

PR #577, opened 2022-03-07
added has completed and has not completed

### LINKED ISSUE #574: [NTR] has_completed
[TICKET IN PROGRESS]

Currently CL uses a custom http://purl.obolibrary.org/obo/cl#has_completed 
This seems to refer to cells that have completed some GO biological process 
eg: memory T cell has_completed some 'memory T cell differentiation'
Would be good to move this to RO. 

rdfs:label: has completed
definition: a relation between a material entity and a process in which the material entity exists, at least in part, due to the completion of the process. 
Example of usage: memory T cell has_completed some 'memory T cell differentiation'
Inverse of: has not completed 

rdfs:label: has not completed 
definition:

shawntanzk:
Fixes #574


dosumis:
CC @addiehl

cmungall:
you will need to resolve conflicts

commens on the relation in the ticket...

cmungall:
What is the status of this - I see a lot of discussion on the issue but not clear if this is ready to be merged?

shawntanzk:
> What is the status of this - I see a lot of discussion on the issue but not clear if this is ready to be merged?

lol, good question. From reading I think theses are the possible solutions:

1. has_completed -> output_of (this can be done with all the options below)
2. make an annotation property for has_not_completed that expands to have the negative (not 100% sure how to handle this personally)
3. use capable_of instead of has_not_completed (downside is that we have to make sure individual changes make sense, and reasoner won't work well here) 

I've converted this to draft as I think we are using this more as a discussion point, happy to implement stuff once we have a clear direction (thought I guess that is the difficult part)

PR #581, opened 2022-03-08
Adding RO templates, including for interaction patterns. Addresses #521

cmungall:
Addresses:

- #521


cmungall:
Add minimal metadata fields:

 - attribution (ORCID): URI dcterms:contributor

PR #584, opened 2022-03-08
Make overlaps symmetric

### LINKED ISSUE #580: overlaps has lost its symmetry axiom
not sure how we have not noticed this for so long

overlaps is clearly symmetric by the basic axioms of mereology

matentzn:
fix #580 

I tried looking very far back in time, and I cannot find any evidence of overlaps ever being symmetric..

Is there any ontology that has plenty of overlaps relations to test this with?

cc @dosumis who I think uses overlaps a lot


cmungall:
> Is there any ontology that has plenty of overlaps relations to test this with?

I suspect it is rarely asserted, and even more rarely asserted used at the abox level where symmetry is most useful

if you do infer it from has-part/part-of at the abox level you will infer symmetry, so you would really only notice this if you assert overlaps at the abox level which is likely rare!

PR #583, opened 2022-03-08
Add "myristoylates" and helper property "is myristoyltransferase activity"

cmungall:
- Adding RO templates, including for interaction patterns. Addresses #521
- docs
- docs
- docs
- staging
- stage
- Add "myristoylates" and helper property "is myristoyltransferase activity".

PR #589, opened 2022-03-08
Add additional PTMs

cthoyt:
As a follow-up to #581, #583, and #586, this PR adds several more PTM relationships. It also adds a slot for defining opposites.

Alternatively, this structure might be better for collapsing some stuff in the `ObjectPropertyDefinedByInteractionProcess` which needs two separate entries for addition and removal right now: 

```yaml
- conforms_to: ObjectPropertyDefinedByInteractionProcess
  uses_rolified_property:
    id: RO:0018001
    label: is myristoyltransferase activity
    rolification_of: GO:0019107
  definition: >-
    A molecularly-interacts-with relationship between two entities,
    where the subject catalyzes a myristoylation activity that takes the object as input
  contributor: orcid:0000-0003-4423-4370
  subproperty_of: RO:0002436  ## molecularly interacts with
  addition:
    id: RO:0018002
    label: myristoylates
    inverse_of:
      id: RO:0018003
      label: myristoylated by
      definition: inverse of demyristoylates
  removal:
    id: RO:0018004
    label: demyristoylates
    inverse_of:
      id: RO:0018005
      label: demyristoylated by
      definition: inverse of demyristoylates
  subject_to_process_property: RO:0002215  ## capable of
  process_to_object_property: RO:0002400   ## has direct input
```


src/patterns/ro-interaction-relations-src.yaml
```diff
@@ -22,8 +22,13 @@
 
 # new one proposed in https://github.com/oborel/obo-relations/pull/522
 # TODO: allow contributor to 'flow through' all named entities by default
-- id: RO:0018002
-  conforms_to: ObjectPropertyDefinedByInteractionProcess
+- conforms_to: ObjectPropertyDefinedByInteractionProcess
+  uses_rolified_property:
+    id: RO:0018001
```
cthoyt, line 27: note I re-ordered this so the IDs are monotonically increasing

cthoyt:
Seemed to not be an issue but I got the following warning from funowl @hsolbrig

```
/Users/cthoyt/.virtualenvs/obo/lib/python3.10/site-packages/funowl/literals.py:180: UserWarning: Code: f is not defined in namespace XSD
  l = g.value(XSD.f, RDF.type)
```

src/patterns/ro-interaction-relations-src.yaml
cmungall: great! We should probably document this. I think for our own sanity it makes sense for people to allocate in consecutive blocks of N, and even have an ID range dedicated to pattern based additions

cmungall (approved):
another thought: there probably isn't much value in continuing to check in the intermediate files, especially if we are seeding rather than compiling components. I did this to make things explicit at the outset but we may want to drop as this will increase the chance of conflicts

PR #593, opened 2022-03-14
Submission of ADO and EPIO related relations

BideZ:
Totally 15 relations from ADO and EPIO are to be submitted.


matentzn (approved):
@BideZ I assigned the range to you, and gave you 100 instead of just 15, so you can add more relations in the future.

PR #595, opened 2022-03-16
Submission of relations(related to ADO and EPIO)

BideZ:
Here are 15 object properties related to ADO and EPIO that we would like to submit to RO.

### DIFF
--- a/src/templates/new_relations.tsv
+++ b/src/templates/new_relations.tsv
@@ -1,16 +1,29 @@
 relation id (required)	type	label (required)	definition (required)	parent relation	creator ORCID (required)	Date created (required)	relation characteristics	domain	range	inverse property	example of use	synonyms	curator/usage notes
 ID	TYPE	AL rdfs:label@en	A IAO:0000115	SP %	AI dc:contributor	AT dc:created^^xsd:dateTime	CHARACTERISTIC SPLIT=|	DOMAIN	RANGE	IP %	A rdfs:comment	A oboInOwl:hasExactSynonym SPLIT=|	A IAO:0000116
-20001 TYPE  isAnatomicalEntityFor A relation describing an anatomical featureof the referenced object with a material entity   AI dc:contributor0000-0002-6571-0033 AT dc:created^^xsd:16.03.2022 
-20002 TYPE  isBrainConnectivityFor  The subject describes the estimated relation between brain regions in a given condition (object). 
-20003 TYPE  isCellularProcessFor  The subject describes any process that is carried out at the cellular level, but not necessarily restricted to a single cell for the object.  
-20004 TYPE  isClassificationFor The subject is a subcategory of the object with certain criteria. 0000-0002-6571-0033 16.03.2022
-20005 TYPE  isDiagnosisFor  The subject is a technique carried out in a clinical assessment for evaluation of a condition (object). 
-20006 TYPE  isEtiologyFor The subject is a cause or result in the pathogenesis of a condition (object). 
-20007 TYPE  isImitatorFor The subject shares some similarities with a certain condition (object) and can be misdiagnosed as this condition. 
-20008 TYPE  isRiskFactorFor The subject possesses characteristics that lead to an increased risk in resulting in a certain phenotype or condition (object).
-20009 TYPE  isSeizureClassificationFor  The subject assignes the object a seizure class.
-20010 TYPE  isSignAndSymptomFor The subject describes a sign/symptom for the object.
-20011 TYPE  isSyndromeFor The subject describes a syndrome correlated to a condition (object).  
-20012 TYPE  isTreatmentFor  The subject describes a clinically approved drug or therapy for a condition (object). 
-20013 TYPE  isEndpointFor The subject whose level or measurement confirms an onset or an existance of a condition (object).
-20014 TYPE  isModelFor  The subject serves as an in vitro/in vivo model for the study of a certain phenotype (object) during pre-clinical phase.
+RO:0020001 owl:ObjectProperty  is anatomical entity for A relation describing an anatomical feature of the referenced object with a material entity   AI dc:contributor0000-0002-6571-0033 AT dc:created^^xsd:16.03.2022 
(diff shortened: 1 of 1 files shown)


src/templates/new_relations.tsv
```diff
@@ -1,2 +1,16 @@
 relation id (required)	type	label (required)	definition (required)	parent relation	creator ORCID (required)	Date created (required)	relation characteristics	domain	range	inverse property	example of use	synonyms	curator/usage notes
 ID	TYPE	AL rdfs:label@en	A IAO:0000115	SP %	AI dc:contributor	AT dc:created^^xsd:dateTime	CHARACTERISTIC SPLIT=|	DOMAIN	RANGE	IP %	A rdfs:comment	A oboInOwl:hasExactSynonym SPLIT=|	A IAO:0000116
+20001 TYPE  isAnatomicalEntityFor A relation describing an anatomical featureof the referenced object with a material entity   AI dc:contributor0000-0002-6571-0033 AT dc:created^^xsd:16.03.2022
```
matentzn, line 3: You need to add `RO:` prefix for all IDs, and pad the ID to have exactly 7 digits, like:

```
RO:0020001
```

src/templates/new_relations.tsv
```diff
@@ -1,2 +1,16 @@
 relation id (required)	type	label (required)	definition (required)	parent relation	creator ORCID (required)	Date created (required)	relation characteristics	domain	range	inverse property	example of use	synonyms	curator/usage notes
 ID	TYPE	AL rdfs:label@en	A IAO:0000115	SP %	AI dc:contributor	AT dc:created^^xsd:dateTime	CHARACTERISTIC SPLIT=|	DOMAIN	RANGE	IP %	A rdfs:comment	A oboInOwl:hasExactSynonym SPLIT=|	A IAO:0000116
+20001 TYPE  isAnatomicalEntityFor A relation describing an anatomical featureof the referenced object with a material entity   AI dc:contributor0000-0002-6571-0033 AT dc:created^^xsd:16.03.2022 
+20002 TYPE  isBrainConnectivityFor  The subject describes the estimated relation between brain regions in a given condition (object). 
+20003 TYPE  isCellularProcessFor  The subject describes any process that is carried out at the cellular level, but not necessarily restricted to a single cell for the object.
```
matentzn, line 5: The label for all relations should be all lower case and with spaces. like 

"is cellular process for".

src/templates/new_relations.tsv
```diff
@@ -1,2 +1,16 @@
...
 ID	TYPE	AL rdfs:label@en	A IAO:0000115	SP %	AI dc:contributor	AT dc:created^^xsd:dateTime	CHARACTERISTIC SPLIT=|	DOMAIN	RANGE	IP %	A rdfs:comment	A oboInOwl:hasExactSynonym SPLIT=|	A IAO:0000116
+20001 TYPE  isAnatomicalEntityFor A relation describing an anatomical featureof the referenced object with a material entity   AI dc:contributor0000-0002-6571-0033 AT dc:created^^xsd:16.03.2022 
+20002 TYPE  isBrainConnectivityFor  The subject describes the estimated relation between brain regions in a given condition (object). 
+20003 TYPE  isCellularProcessFor  The subject describes any process that is carried out at the cellular level, but not necessarily restricted to a single cell for the object.  
+20004 TYPE  isClassificationFor The subject is a subcategory of the object with certain criteria. 0000-0002-6571-0033 16.03.2022
+20005 TYPE  isDiagnosisFor  The subject is a technique carried out in a clinical assessment for evaluation of a condition (object).
```
matentzn, line 7: every isXFor should have a corresponding `has X` relation. So you should also add another `has diagnosis` relation here.

src/templates/new_relations.tsv
```diff
@@ -1,2 +1,16 @@
...
+20001 TYPE  isAnatomicalEntityFor A relation describing an anatomical featureof the referenced object with a material entity   AI dc:contributor0000-0002-6571-0033 AT dc:created^^xsd:16.03.2022 
+20002 TYPE  isBrainConnectivityFor  The subject describes the estimated relation between brain regions in a given condition (object). 
+20003 TYPE  isCellularProcessFor  The subject describes any process that is carried out at the cellular level, but not necessarily restricted to a single cell for the object.  
+20004 TYPE  isClassificationFor The subject is a subcategory of the object with certain criteria. 0000-0002-6571-0033 16.03.2022
+20005 TYPE  isDiagnosisFor  The subject is a technique carried out in a clinical assessment for evaluation of a condition (object). 
+20006 TYPE  isEtiologyFor The subject is a cause or result in the pathogenesis of a condition (object). 
+20007 TYPE  isImitatorFor The subject shares some similarities with a certain condition (object) and can be misdiagnosed as this condition.
```
matentzn, line 9: TYPE should be `owl:ObjectProperty` for all relations.

matentzn (changes requested):
First round of feedback. Thanks!

BideZ:
@matentzn Thanks for the feedback. I have committed the changes about the aforementioned issues. I added hasX for all isXFor relations except 'is model for' because 'has model' is already there in the relation Ontology. Besides, I defined all the hasX relations in a fashion consistent with 'has model' in RO (Inverse of is-X-For ). I am looking forward to your reply! Thank you.

matentzn:
@BideZ Technically this looks good to me. Now the hard part - convincing the RO people which relations are really needed.  Unfortunately, this is no longer my area, but I am happy for your first attempt to interact with RO. Note that getting stuff into these "upper level" ontologies can be quite a lengthy process.. I hope someone from the RO team will pick this up and help you with some concrete steps to move this forward.

cmungall:
apologies I am just seeing this for the first time now

While it's good to batch requests I think a lot of these need to be discussed and things might get lost in a single over-arching issue. Overall I would say:

- there appears to be a lot of duplication with existing relations, these could be accomodated with synonym requests (e.g. adding sign or symptom onto existing phenotype relations)
- there are no domain/range constraints. While this is a challenge in general due to lack of agreement on TLOs there are some cases where obvious D/R axioms could be made

Issue #601, opened 2022-04-21
Some terms in ro.obo file have text strings for ids


vanaukenk:
Hi,

In parsing the ro.obo file for WormBase, we've seen that some RO terms have a text string for an id in the ro.obo file.  Here's an example:

[Typedef]
id: results_in_acquisition_of_features_of
name: results in acquisition of features of
def: "The relationship that links a specified entity with the process that results in an unspecified entity acquiring the features and characteristics of the specified entity" []
subset: http://purl.obolibrary.org/obo/valid_for_go_annotation_extension
subset: http://purl.obolibrary.org/obo/valid_for_go_ontology
subset: http://purl.obolibrary.org/obo/valid_for_gocam
property_value: IAO:0000112 "an annotation of gene X to cell differentiation with results_in_maturation_of CL:0000057 (fibroblast) means that at the end of the process the input cell that did not have features of a fibroblast, now has the features of a fibroblast." xsd:string
property_value: IAO:0000117 "Chris Mungall" xsd:string
property_value: IAO:0000119 GOC:mtg_berkeley_2013 xsd:string
range: CARO:0000003 ! connected anatomical structure
is_a: results_in_developmental_progression_of ! results in developmental progression of
is_a: RO:0040036 ! results in changes to anatomical or cellular structure

In ro.owl, this term name is associated with RO:0002315

Is it intentional to use the term name for an id in the ro.obo file and, if not, can this be fixed and a new ro.obo file generated?

Thanks!

matentzn:
Wow, this is true for 7 relations in ro.obo:

[Typedef]
id: results_in_acquisition_of_features_of
name: results in acquisition of features of
def: "The relationship that links a specified entity with the process that results in an unspecified entity acquiring the features and characteristics of the specified entity" []

[Typedef]
id: results_in_development_of
name: results in development of
def: "p 'results in development of' c if and only if p is a developmental process and p results in the state of c changing from its initial state as a primordium or anlage through its mature state and to its final state." []

[Typedef]
id: results_in_developmental_progression_of
name: results in developmental progression of
def: "p results in the developmental progression of s iff p is a developmental process and s is an anatomical structure and p causes s to undergo a change in state at some point along its natural developmental cycle (this cycle starts with its formation, through the mature structure, and ends with its loss)." []

[Typedef]
id: results_in_disappearance_of
name: results in disappearance of

[Typedef]
id: results_in_formation_of
name: results in formation of

[Typedef]
id: results_in_maturation_of
[...]

PR #603, opened 2022-04-25
2022-04-25 Release

shawntanzk:
Fixes https://github.com/oborel/obo-relations/issues/601


shawntanzk:
I've checked and obo format looks alright once I removed shorthand - need to go through diff for release still

matentzn (approved):

PR #605, opened 2022-05-02
change label 'utilizes' to 'device utilizes material'

wdduncan:


wdduncan:
Fixes #497

@cmungall @linikujp Mainly, this changes the label `utilizes` to `device utilizes material`. I also used Asiyah's ORCID as the value of `term editor`.

cmungall (changes requested):
merge after adding utilizes as a synonym (broad?)

PR #610, opened 2022-06-07
Add symmetric to continuous with

anitacaron:
Related to https://github.com/obophenotype/uberon/issues/2453


matentzn:
No need to assign me to such issues - these must be dealt with by @dosumis and @cmungall which are the gatekeepers of RO semantics!

balhoff (approved):
Approved at RO meeting 2022-06-28.

PR #611, opened 2022-06-09
Add classification to `simultaneous with`

anitacaron:
Related to https://github.com/obophenotype/uberon/issues/2453


balhoff:
Discussion at RO meeting 2022-06-28: we should have definitions for 'simultaneous with', 'ends' and 'starts' in order to decide how they are related.

dosumis:
The comment has a proposed def:

t1 simultaneous_with t2 iff:= t1 before_or_simultaneous_with t2 and not (t1 before t2).  

This is certainly not ideal, but I think it indicates that this is not meant to be used between processes/intervals but between points in time - where the points in time we are interested in our those that define the start and end points of an occurrent.  Defined in this way, the term should not be used in Uberon (where this request originates).  I think what's needed for Uberon is probably the Allen 'equals' relation:

image
http://www.thomasalspaugh.org/pub/fnd/allen.html. 

We don't seem to have that at present.

To be consistent with other defs, this should be something like this:  

`x {name_TBD} y iff ω(x) = ω(y) and ω(α ) = ω(α), where α is a function that maps a process to a start point, and ω is a function that maps a process to an end point and '=' indicates the same instance in time.`

(Note - the `=` above is the same as the intent of the current 'simultaneous_with relation)

I think

TBD:
*  do we repurpose simultaneous_with for this?  If we do we should drop the grouping term and the 'before' objectProperty as both of these are also apply between instance in time?  Perhaps the strictly correct thing to do would be to obsolete this term and create a new simultaneous_with using the above def. 

Re - where this should be in the property hierarchy - to be a subproperty of starts and ends, starts and ends should be defined as α(y) = α(x) and ω(y) = ω(x) respectively.

More generally - it would be really awesome if we could find a decent logician to review this whole branch.  We decided years ago to not directly re-use Allen relations, but to use derivable relations that (we believe(d)) are more intuitive for biologists.  For example our 'precedes' is a subproperty of Allen 'meets'.  The general work required is to derive property heirarchy &  chains (or rules) and standardise definitions - using Allen as a starting point.  Here's a rather ancient attempt: https://docs.google.com/document/d/1kBv1ep_9g3sTR-SD3jqzFqhuwo9TPNF-l-9fUDbO6rM/edit?pli=1

shawntanzk:
From tech call

Minimum:
- [ ] obsolete this term and create a new simultaneous_with using the above def. @cmungall is this definition acceptable, and shall we create a new relation? 

Other parts of https://github.com/oborel/obo-relations/pull/611#issuecomment-1173714948 should be opened as a new ticket for longer term modelling solutions

cmungall (approved):

balhoff:
I'm not sure if this is ready to be merged; @cmungall approved but @shawntanzk's comment suggests some changes?

wdduncan:
@dosumis @shawntanzk Can you attend the next RO call so that we can discuss this.

cmungall (changes requested):
I mistakenly approved before. This is actually incorrect because of the `<` condition on starts/ends

anitacaron:
I did the changes. Just to make it clear that this relation is already transitive. This type is in the `temporal-intervals.owl`

cmungall (approved):

PR #615, opened 2022-07-11
Add SPARQL expansion for 'mutually spatially disjoint with’.

balhoff:
Fixes #600.

PR #616, opened 2022-07-11
Add SPARQL expansions for 'never in taxon' and 'present in taxon'

balhoff:
Relies on annotation property added by #615; merge that one first.


dosumis:
Is the explicit negation within EL(++)?  IIRC it can be under some circumstances (on left in GCI?).

balhoff:
@dosumis that negation does work with ELK to a limited extent. I added that axiom to address this situation, which was a loophole preventing property chains working with 'never in taxon': https://github.com/geneontology/go-annotation/issues/3942#issuecomment-940579254

It requires this expansion to the taxon disjoints in order to do anything, though: https://github.com/geneontology/go-ontology/pull/22509/files#diff-3b5490b146d7b78e80eac2850ebac5c8846c4aa28168ff929e04a5fea7c01964R5-R17

cmungall (approved):

PR #649, opened 2022-11-11
tiny PR to remove a broken link in the documentation.

sierra-moxon:


wdduncan:
@sierra-moxon Thanks for catching this!  
The wording on the page is in general kind of odd. The title of the page is "Property chains", then it talks a little bit about property chains, then it closes with "Property chains in RO".   
It me be better to have the main header/title say "Property chains in RO", say a little about what a property chain is, and then close with a few extra examples of property chains.

Thoughts?

wdduncan (approved):

PR #652, opened 2022-11-17
changed term label for directly provides input for

pgaudet:


balhoff (approved):

wdduncan:
I thought the reason for the use of `directly` was to distinguish relations that **immediately** proceed something from those relations that just **proceed** something. Is this correct? Does changing the label blur the distinction?

balhoff:
@pgaudet @vanaukenk @ukemi could you answer Bill's question?

wdduncan:
I am looking at the documentation [here](https://oborel.github.io/obo-relations/direct-and-indirect-relations/#direct-and-indirect-forms-of-relations).

It says:
> the 'direct' prefix means that the two structures or classes should be in direct succession (no intermediates).

Does changing the label invalidate this?

vanaukenk:
Thanks for your input @wdduncan 

We don't actually think 'provides input for' quite fits the pattern in the RO documentation.

We thought about how to uniquely define 'provides input for' vs 'directly provides input for' and came to the conclusion that there really only was one relation, 'provides input for' and that relation describes a direct connection between an upstream occurrent that is the provider of an input used by the downstream occurrent.  Changing the term label shouldn't invalidate this.

We couldn't think of any cases where there is an intervening/intermediate occurrent.

Note also the comment that currently exists on this term:

"This is currently called 'directly provides input for' to be consistent with our terminology where we use 'direct' whenever two occurrents succeed one another directly. We may relabel this simply 'provides input for', as directness is implicit")."

wdduncan:
@vanaukenk Thanks for the explanation. The definition for `provides input for` makes it clear the relation is "immediate":

> p provides input for q iff p is immediately causally upstream of q, and there exists some c such that p has_output c and q has_input c.

I just have slight worry that only developers will be reluctant to follow the RO recommendation I referenced, and use this a justification. But, we'll see what happens ...

wdduncan:
@vanaukenk @balhoff @pgaudet @ukemi How general/granular is this relation intended to be? 

For example, chewing food (a process) provides input digestive process. However, there are intervening processes (e.g., swallowing). So, chewing food cannot "provide input for" digestion.

balhoff:
I guess 'chewing food' [transitively provides input for](http://purl.obolibrary.org/obo/RO_0002414) 'digestion'.

wdduncan:
Good point @balhoff  
Yes, that works, and it points out another deviation from the [RO documentation](https://oborel.github.io/obo-relations/direct-and-indirect-relations/#direct-and-indirect-forms-of-relations). The documentation indicates (to me) that the hierarchy should be:
```
indirectly provides input for
  - directly provides input for
```

I the digestion example was mainly in response to this comment:
> ... came to the conclusion that there really only was one relation, 'provides input for' and that relation describes a direct connection between an upstream occurrent

It would have been nice if you all would have waited until the next RO call before merging.

balhoff:
My apologies, Bill! In the past some of these PRs and requests have just lingered so long that I was trying to be proactive. I am happy to discuss in the next call; I think we should invite some of the GO folks if so.

wdduncan:
You are definitely correct that some PRs have lingered for too long. I am hoping monthly RO meetings may address this. The GO folks are more than welcome to join the call :)

PR #655, opened 2022-11-22
add relations about measurements of characteristics

### LINKED ISSUE #535: NTR 'measurement method of' (previously in AgrO)
Hello,
I'd like to request adding [`measurement method of`](http://purl.obolibrary.org/obo/AGRO_00000253) to RO along with the following details:

`Label`: measurement method of
`Definition`: A relation between a plan specification and a 'specifically dependent continuant', in which the plan specification groups the specifically dependent continuant(s).
`Uses`: [area measurement protocol](http://purl.obolibrary.org/obo/AGRO_00000286), [material area density mesurement protocol](http://purl.obolibrary.org/obo/AGRO_00000242), [pivot lenght measurement protocol](http://purl.obolibrary.org/obo/AGRO_00000282), etc
`ORCID`: https://orcid.org/0000-0002-8461-9745

Can be a discussion if domains and ranges are needed.
[plan specification](http://purl.obolibrary.org/obo/IAO_0000104)
[specifically dependent continuant](http://purl.obolibrary.org/obo/BFO_0000020)

Originally a question in issue #505

@marieALaporte

### LINKED ISSUE #653: NTR: is about
Not really a new term request (I suppose). But #535 made me realize that `is about` is not included as part of RO. Should we be importing it from IAO? May be useful for other "about" relations that may be needed.

### LINKED ISSUE #654: NTR: is characteristic measurement of
Add `is measurement of characteristic` to be more generic than IAO's `is quality measurement of`. E.g., processes can have characteristics, and this would allow the process in the range of the relations.

wdduncan:
- add target to create iao import
- Fixes #653
- add iao_import to ro-edit
- fix: change term editor to dc:contributor with orcid value
- Fixes #535
- add background info in comment
- Fixes #654


jamesaoverton:
Can this please be marked as Draft until the OBI group has a chance to discuss it?

wdduncan:
@jamesaoverton it has been converted to a draft

jamesaoverton:
I'm worried about RO importing from IAO. If 'is about' is needed in RO, can it be moved instead?

For the current use cases, I don't see why anything more than 'is about' is needed, but this PR imports several more IAO terms.

wdduncan:
@jamesaoverton 
> I'm worried about RO importing from IAO. If 'is about' is needed in RO, can it be moved instead?

Moving is an option, but IAO would need to agree to it. I submitted an [issue](https://github.com/information-artifact-ontology/IAO/issues/264) getting there opinion.

> For the current use cases, I don't see why anything more than 'is about' is needed, but this PR imports several more IAO terms.

Given how broad `is about` is, I included the child relations to help differentiate the different senses in which an ICE can "be about" something else.

ddooley:
I think this can be closed without commit?

wdduncan:
The issues that generated this PR have been closed. It was determined the relations were not needed.

wdduncan:
@ddooley Yes ... I've closed it :)

PR #659, opened 2022-12-23
Relax the Range on RO_0002295

shawntanzk:
Change the range of RO to be anatomical entity instead 
Related to https://github.com/obophenotype/cell-ontology/issues/1782 (see CL ticket for reason)
Thanks to @gouttegd & @bvarner-ebi


gouttegd:
The definition should be corrected as well, so that it no longer refers to an _anatomical structure_.

wdduncan:
I am happy to put this on the agenda for the next RO meeting. Can you please hold off merging until then?

cmungall:
I think this is good to merge, once Damien's comments are incorporated
(this could even go in a separate PR)

shawntanzk:
> Damien's comments are incorporated

Already incorporated :) 

> I think this is good to merge

Alright, will merge, no need for taking up time in RO meeting then :) Thanks @cmungall

PR #660, opened 2022-12-23
relax domain/range on regulates relations

shawntanzk:
see https://github.com/obophenotype/cell-ontology/issues/1782#issuecomment-1363995278 for reasoning
Thanks to @gouttegd & @bvarner-ebi

### DIFF
--- a/src/ontology/ro-edit.owl
+++ b/src/ontology/ro-edit.owl
@@ -2320,7 +2320,6 @@ AnnotationAssertion(rdfs:label obo:RO_0002211 "regulates"@en)
 SubObjectPropertyOf(obo:RO_0002211 obo:RO_0002411)
 InverseObjectProperties(obo:RO_0002211 obo:RO_0002334)
 TransitiveObjectProperty(obo:RO_0002211)
-ObjectPropertyDomain(obo:RO_0002211 obo:BFO_0000015)
 ObjectPropertyRange(obo:RO_0002211 obo:BFO_0000015)
 
 # Object Property: obo:RO_0002212 (negatively regulates)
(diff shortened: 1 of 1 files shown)


gouttegd:
I believe the proposed fix is either incomplete, or wrong.

`regulated by` and `regulates` are both (indirect) subclasses of `causal relation between processes` (RO:0002501); the definition of `regulates` (and, by reflection, the definition of `regulated by`) clearly implies that both the subject and the object should be processes:

> p regulates q iff p is causally upstream of q, the execution of p is not constant and varies according to specific conditions, and p influences the rate or magnitude of execution of q due to an effect either on some enabler of q or some enabler of a part of q.

(Note it mentions both “the execution of p” and “the execution of q”, implying both _p_ and _q_ are about things that happen – i.e. processes.)

So the range and domain restrictions on these relations are consistent with both their definitions and their positions in the hierarchy. If we remove those restrictions to allow these relations to be used between a process and something else, then the definition of _regulates_ should be amended accordingly, and (perhaps most importantly) both relations should be re-classified so that they are no longer under `causal relation between processes`.

I tend to think both relations are actually correct (including in their range/domain restrictions), and that the problem highlighted in obophenotype/cell-ontology#1782 should be fixed on the GO side, by using another relation that `regulated by` which to me seems clearly _not_ intended to be used the way GO has used it.

shawntanzk:
Happy to close this, I didn't think about hierarchy and all but you are right. Would be good to have a comment or smth to ensure that it isn't use wrongly again :) thanks

dosumis:
Should be clear from the definition as well as the hierarchy.  Not sure how else this can be flagged.  I think the important issue here is why this was missed by GO QC.  Maybe range still not being checked as using older ELK?

GO should fix by changing to regulated by 'Ca2+ binding'.

PR #661, opened 2023-01-03
Relax range and domain on RO:0002256

shawntanzk:
see https://github.com/obophenotype/uberon/issues/2749

relation: 'developmentally induced by'

### DIFF
--- a/src/ontology/ro-edit.owl
+++ b/src/ontology/ro-edit.owl
@@ -2705,7 +2705,7 @@ SubObjectPropertyOf(obo:RO_0002255 obo:RO_0002385)
 # Object Property: obo:RO_0002256 (developmentally induced by)
 
 AnnotationAssertion(obo:IAO_0000114 obo:RO_0002256 obo:IAO_0000125)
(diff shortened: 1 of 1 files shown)


dosumis:
Please use 'material anatomical entity' rather than just 'anatomical entity'. I don't think it makes sense for an 'immaterial anatomical entity' to induce anything.

In CARO this is http://purl.obolibrary.org/obo/CARO_0000006 - but @cmungall should we be switching these to Uberon?

shawntanzk:
> Please use 'material anatomical entity' rather than just 'anatomical entity'. I don't think it makes sense for an 'immaterial anatomical entity' to induce anything.

Ah you're right didn't think about immaterial anatomical entities, will make the change :)

dosumis (approved):

shawntanzk:
In RO call we decided to handle this asynchronously. I see that David has given the thumbs up so I will go ahead and merge this, please do open another ticket if you have concerns (relaxing wont break anything, so we can always tighten if need be without any pipelines breaking). 
Sorry for the rush, planning to make a RO release soon to fix unsats :)

PR #665, opened 2023-01-06
Replace nonsense "has term editor" annotation assertion with "has exact synonym" in RO:0002207

cthoyt:
`"has developmental precursor"` is clearly not the name of a person, so it doesn't belong using the `has term editor` property from IAO. Maybe it's a synonym? @cmungall @dosumis since you're marked as the editors can you clarify how to address this?


cmungall:
This kind of annotation property slip up is common in protege. It should be
a synonym, I think exact

cthoyt:
@cmungall done

matentzn (approved):
Gr8, thanks

PR #685, opened 2023-02-22
Restoring annotations.owl module but in deprecated state

### LINKED ISSUE #683: Suggest restoring `annotations.owl`
OBI uses <http://purl.obolibrary.org/obo/ro/releases/2015-10-07/core.owl>, which imports <http://purl.obolibrary.org/obo/ro/annotations.owl>, which no longer resolves (404) after #563. The underlying problem is that the old releases such as 2015-10-07 used imports without versions, as mentioned in #154. That old release also imports `bfo-classes-minimal.owl`, but that file was not dropped by #563.

The simplest solution would be to just restore `annotations.owl` at the top level of this repo.

cmungall:
Fixes #683

### DIFF
--- a/annotations.owl
+++ b/annotations.owl
@@ -1,0 +1,328 @@
+<?xml version="1.0"?>
+<rdf:RDF xmlns="http://purl.obolibrary.org/obo/bfo.owl#"
+     xml:base="http://purl.obolibrary.org/obo/bfo.owl"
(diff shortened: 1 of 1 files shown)


matentzn (approved):
The more elegant solution would have been to add a purl config entry redirecting to the last tag the file existed but this is also fine.

PR #492, opened 2021-09-08
RO migration to ODK 1.4

gouttegd:
Fixes #563

- chemicals where merged into EDIT

PR #688, opened 2023-02-27
Import ORCID Instance Ontology

### LINKED ISSUE #666: Import ORCIDIO
Following #664, there are now a lot more canonical ORCID URIs. while this has the benefit of making the provenance information in RO structured and actionable, it comes with the cost of readability. ORCIDIO (https://github.com/cthoyt/orcidio) is an ontology that declares canonical ORCID URIs as named individuals and adds labels, so you can see beautifully annotated term editors, creators, contributors, etc.

What's the best way to import ORCIDO? Some possibilities:

1. During release time (this doesn't benefit people who are working on the ro-edit.owl
2. Directly into ro-edit.owl. How should this get maintained? Could ODK do this for us?

cthoyt:
Closes #666

### DIFF
--- a/src/ontology/ro-odk.yaml
+++ b/src/ontology/ro-odk.yaml
@@ -40,6 +40,8 @@ import_group:
     - id: other
       module_type: custom
       mirror_type: no_mirror
+    - id: orcidio
+      module_type: filter
 subset_group:
   products:
     - id: ro-interaction


cmungall (approved):

cthoyt:
This still needs a bit of magic from nico before merging

matentzn:
This works, but at least in my protege the label is not rendered. I don't really now how to set Protege to render labels correctly.

src/ontology/components/temporal-intervals.owl
```diff
@@ -10,7 +10,6 @@
      xmlns:foaf="http://xmlns.com/foaf/0.1/"
      xmlns:dc="http://purl.org/dc/elements/1.1/">
     <owl:Ontology rdf:about="http://purl.obolibrary.org/obo/ro/components/temporal-intervals.owl">
-        <owl:imports rdf:resource="http://purl.obolibrary.org/obo/ro/core.owl"/>
```
matentzn, line 13: This is an oversight from the #492 PR sorry about that. Couldnt refresh imports w/o fixing this.

src/ontology/ro-odk.yaml
```diff
@@ -19,6 +19,11 @@ namespaces:
   - http://purl.obolibrary.org/obo/BFO_0000051
   - http://purl.obolibrary.org/obo/BFO_0000060
 import_group:
+  annotation_properties:
+        - rdfs:label
+        - http://purl.org/dc/elements/1.1/description
```
cthoyt, line 24: Can we use curies for these?

cthoyt:
@matentzn this is great. Given you’ve put in the effort to do this here, what do you think next steps are to do this in other ontologies? Can we make a specific guide? Or can we add a command to set this up to the odk itself?

cthoyt:
@matentzn note based on your comment I have updated all of the DC URIs to use DCTERMS in https://github.com/cthoyt/orcidio/commit/08da71ba8875d85e51cc4b962da978d9389d0eff, this might mean you should revise this PR a bit

matentzn (approved):

PR #692, opened 2023-02-28
Replace all uses of DCE with DC and add violation check

### LINKED ISSUE #691: Deprecate use of DC elements namespace
Make sure no uses of dce happens in RO, we should only use `dc:` (where dc is the "terms" namespace).

cthoyt:
Closes #691

This PR does the following:

1. Replaces all remaining uses of `DCE` with `DC`
2. Removes the `DCE` prefix from the definitions and clean up corresponding declarations
3. Adds a violation check to make sure no predicates use the URI prefix for `DCE`

I have just been educated today by Anita on how to set up the violation check - I'll need someone to explain the best way to check this (does the PR itself run these checks?)


cmungall (approved):

matentzn (approved):
We are now getting to see how useful qc profile centralisation would be. @StroemPhi

matentzn:
Ready to merge after conflict resolution

cthoyt:
@matentzn merge/conflict resolution is done. Next steps:

1. Upstream these QC profiles into ODK
2. How do we ensure that sparql_test is run as part of QC for pull requests?

matentzn:
> Upstream these QC profiles into ODK

PR here: https://github.com/INCATools/ontology-development-kit/tree/master/template/src/sparql

> How do we ensure that sparql_test is run as part of QC for pull requests?

It always is: 

https://github.com/oborel/obo-relations/pull/692/checks#step:4:26

PR #696, opened 2023-03-17
NTR "is matched small molecular pair with"

cthoyt:
Chemicals A and B are a matched (small) molecular pair if their chemical structures define by a single, relatively small, well-defined structural modification. In the following example from Wikipedia, these two moleculers are MMPs, with the yellow part being the variable region.

[A depiction of a matched molecular pair](https://en.wikipedia.org/wiki/Matched_molecular_pair_analysis#/media/File:A_MMP.png)

## Why this is useful

Tools like [mmpdb](https://github.com/rdkit/mmpdb) can be used to generate MMP relationships on datasets like ChEMBL in bulk. These can be included in knowledge graphs [useful for drug discovery](https://www.sciencedirect.com/science/article/pii/S1359644613000937), as MMPs are a discrete alternative to including relationships based on chemical similarity (e.g., cosine similarity) using some featurization (e.g., MACCS keys, ECFP-4) and cutoff.

## Properties of this relationship

- This relationship is symmetric, meaning if A is a matched molecular pair with B iff B is a matched molecular pair with A.
- This relationship is not transitive, meaning that A is a MMP with B and B is a MMP with C, then A is not necessarily an MMP with C.
- Its domain and range are both chemicals (material entities)
- This term falls under the "chemical relationship" (RO_0018030) object property

## Summary of Discussion

On the RO call on March 28th, 2023, there was the following feedback:
1. The label might be misleading in contexts when "molecule" can refer to peptides or other macromolecular entities like proteins. While in the cheminformatics domain, it's standard that MMP corresponds to small molecules, it is helpful to update the label to be more specifi. This was done in 1daf181. However, we can also keep in mind that this relation falls in the "chemical relationship" hierarchy and is therefore implicitly only for small molecules
2. There was some discussion about additional metadata models that might accompany a relationship in a KG like "A is an MMP with B" such as "A and B share the parent structure C". 
    
    These could appear as annotations in an ontology setting or as edge metadata in a KG scenario. For example, "A is an MMP with B" implies that there is some C such that A RO:0018040 (has parent hydride) C and B RO:0018040 C. This relation is part of the proposal in #698.


handemcginty:
Just to clarify, we talked about whether we need to look at some modeling-related questions regarding the differences in molecular structure and how it could be captured.

PR #662, opened 2023-01-03
NTR: Add conjugate acid and base relations

cthoyt:
This PR closes #643 by adding five relationships:

| CURIE | Label | Reason |
|-------|------|------|
| RO:0018034 | is protonated form of | Transitive relationship, mirrors http://purl.obolibrary.org/obo/chebi#is_conjugate_acid_of |
| RO:0018033 | is deprotonated form of | Transitive relationship, mirrors http://purl.obolibrary.org/obo/chebi#is_conjugate_base_of |
| RO:0018032 | is direct conjugate acid of | Corresponds to a one proton change, not transitive. |
| RO:0018031 | is direct conjugate base of | Corresponds to a one proton change, not transitive. | 
| RO:0018030 | chemical relationship | Groups the previous two relations, and can be used to group future relations ported out of [the ChEBI handbook](https://docs.google.com/document/d/1_w-DwBdCCOh1gMeeP6yqGzcnkpbHYOa3AGSODe5epcg/edit#), similarly to how [disease relationship (RO:0040035)](http://purl.obolibrary.org/obo/RO_0040035) groups various disease relationships. If this is too general, we could rename it to `in protonation relationship with` |

This PR does **not** port the `oboInOwl:is_cyclic` and `oboInOwl:is_transitive` that were originally in the ChEBI RDF based on suggestions from Jim.

This PR uses the design pattern described in https://oborel.github.io/obo-relations/direct-and-indirect-relations/

PR #698, opened 2023-03-19
Add remaining 5 ChEBI relations

### LINKED ISSUE #697: Port remaining ChEBI relations to RO
Following #643, there are a few relations from ChEBI that would benefit from inclusion in RO (also as part of a more general effort to get rid of hash relations across the OBO ecosystem):

- has functional parent
- has parent hydride
- is enantiomer of
- is substituent group of
- is tautomer of

These are described in [this internal Google doc for ChEBI annotators](https://docs.google.com/document/d/1EZHaOEl-iPZPbqD_GRyIetDoA4U8U2SPr9Qk6RKVRc0/edit#heading=h.o8b3r6la80rf), but some of the descriptions are pretty difficult to decipher

cthoyt:
Closes #697 

This PR adds five new relationships corresponding to relationships in the ChEBI ontology as a follow-up to #662, where we added two relationships from ChEBI related to conjugate acids and bases.

| Identifier | Label | Description | ChEBI |
|-----------|--------|------------|------|
| RO:0018038 | has functional parent | Chemical A has functional parent Chemical B if there is chemical transformation through which chemical B can be produced from chemical A. <br><br> For example, the relationship between a salt and a freebased compound is a "has functional parent" relationship. | http://purl.obolibrary.org/obo/chebi#has_functional_parent |
| RO:0018040 | has parent hydride | Chemical A has parent hydride Chemical B if there exists a molecular graphical transformation where functional groups on A are replaced with hydrogens in order to yield B. | http://purl.obolibrary.org/obo/chebi#has_parent_hydride |
| RO:0018039 | is enantiomer of | Chemicals A and B are enantiomers if they share the same molecular graph except the change of the configuration of substituents around exactly one chiral center. <br><br> A chemical with no chiral centers can not have an enantiomer. A chemical with multiple chiral centers can have multiple enantiomers, but its enantiomers are not themselves enantiomers (they are diastereomers). | http://purl.obolibrary.org/obo/chebi#is_enantiomer_of |
| RO:0018037 | is substitutent group from | Group A is a substituent group from Chemical B if A represents the functional part of A and includes information about where it is connected. A is not itself a chemical with a fully formed chemical graph, but is rather a partial graph with one or more connection points that can be used to attach to another chemical graph, typically as a functionalization. | http://purl.obolibrary.org/obo/chebi#is_substituent_group_from |
| RO:0018036 | is tautomer of | Two chemicals are tautomers if they can be readily interconverted. <br><br> This commonly refers to prototropy in which a hydrogen's position is changed, such as between ketones and enols. This is also often observed in heterocyclic rings, e.g., ones containing nitrogens and/or have aryl functional groups containing heteroatoms. | http://purl.obolibrary.org/obo/chebi#is_tautomer_of |

I found it pretty difficult to write good definitions of each of these relationships based on the [ChEBI annotator guide](https://docs.google.com/document/d/1EZHaOEl-iPZPbqD_GRyIetDoA4U8U2SPr9Qk6RKVRc0/edit) because the sections on each of these first gave a really abstract definition (not specific enough), then an additional definition full of jargon (too specific), so I did my best to write human-usable definitions. These are assuming some chemistry knowledge, if desired, I can make them more accessible for a general audience (e.g., explaining the referenced chemistry concepts in full)

Each of the following has:

1. A parent relationship to "chemical relationship"
1. Reference to original issue
2. Example from ChEBI
3. Attribution / date of creation
4. Xref to equivalent ChEBI relationship

## Discussion on March 28, 2023 RO Meeting

- @bpeters42 mused whether importing them just because they're in ChEBI is a good idea (i.e., they still should be considered by someone who knows the domain well). He might have some other specific follow-up. In general I (charlie) think that each of these are valuable relations that are well-defined.

### DIFF
--- a/src/ontology/ro-edit.owl
+++ b/src/ontology/ro-edit.owl
@@ -6607,6 +6607,7 @@ SymmetricObjectProperty(obo:RO_0018036)
 # Object Property: obo:RO_0018037 (is substitutent group from)
 
 AnnotationAssertion(Annotation(rdfs:seeAlso obo:CHEBI_30795) Annotation(rdfs:seeAlso obo:CHEBI_58957) obo:IAO_0000112 obo:RO_0018037 "carboxylatoacetyl group (CHEBI:58957) is substituent group from malonate(1-) (CHEBI:30795)")
+AnnotationAssertion(obo:IAO_0000115 obo:RO_0018037 "Group A is a substituent group from Chemical B if A represents the functional part of A and includes information about where it is connected. A is not itself a chemical with a fully formed chemical graph, but is rather a partial graph with one or more connection points that can be used to attach to another chemical graph, typically as a functionalization.")
 AnnotationAssertion(<http://purl.org/dc/elements/1.1/creator> obo:RO_0018037 <https://orcid.org/0000-0003-4423-4370>)
 AnnotationAssertion(<http://purl.org/dc/elements/1.1/date> obo:RO_0018037 "2023-03-18T23:49:31Z"^^xsd:dateTime)
 AnnotationAssertion(oboInOwl:hasDbXref obo:RO_0018037 <http://purl.obolibrary.org/obo/chebi#is_substituent_group_from>)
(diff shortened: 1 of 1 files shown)


cthoyt:
@bpeters42 did you want to comment on this PR as a follow-up to discussion yesterday?

bpeters42:
My musings were more about Chebi in general, and how we are locking us into a project that is not responsive. But on the other hand, there are plenty of worse relationships in RO, and this PR is consistent with what we are aiming to do in terms of moving relationships out of ontologies. So I approve.

CarMoreno:
Thank you for this guys 👍

Issue #107, opened 2016-04-08
RO core imports minimal classes


mark-jensen:
Users of BFO are being instructed to import the classes only version of BFO found [here](http://purl.obolibrary.org/obo/bfo.owl)

and the RO core found [here](http://purl.obolibrary.org/obo/ro/core.owl)

Due to RO core having a minimal subset of BFO classes, this leads to duplication of some labels, defintions and so forth. 

Why does RO core need to import the BFo classes? Any way to resolve this?

@cmungall @jamesaoverton @zhengj2007

cmungall:
They're required for domain and range constraints, and other axioms.

I guess we could refactor further, split into a 

 1. bridging axiom module, 
 2. an import module, 
 3. and a core-core

So BFO could bring in 1 and 3 and ditch 2.

But this kind of thing happens all over the place and we have no good general solution. It's far worse for more dynamic ontologies. See https://github.com/ontodev/robot/issues/38

BFO could also just make its own RO core mirror e.g. as bfo/imports/ro_import.owl, but the modularization may not be straightforward see #106

zhengj2007:
Hi Mark,

If the BFO classes are defined exactly in both BFO 2.0 and RO, it won't 
show duplicated labels, definitions, etc. when you view the ontology in 
Protege or after merged them in one OWL file. However, I found the same 
issues since sometime the labels, definitions provided as English in one 
OWL file but String in another OWL. If we can keep them same, the issue 
can be solved even RO import BFO.

Jie

nlharris:
What is the status of this?

nlharris:
Can this be closed?

gouttegd:
The problem is still very much present.

RO contains (in the `bfo-classes-minimal.owl` component) definitions for some BFO-owned classes that conflict with the definitions found in BFO itself. I guess those definitions _were_ aligned with the BFO definitions at some point, but over time BFO definitions have been updated and those changes have not been reflected in `bfo-classes-minimal.owl` (that component seems to be manually maintained).

This causes issues as soon as one tries to import both BFO and RO in the same ontology.

There would be several approaches to fix the problem:

1) Replace the manually maintained `bfo-classes-minimal.owl` component (along with the `bfo-axioms.owl` component, which contains disjointness axioms) by a proper ODK-style import module, and get the BFO classes needed in RO from that module. This would be the best solution IMHO.

[...]

PR #705, opened 2023-03-29
NTR: owns; is owned by

wdduncan:
Fixes #674

PR #707, opened 2023-03-31
Import BFO

gouttegd:
This PR replaces the manually maintained BFO components (`bfo-classes-mininal.owl` and `bfo-axioms.owl`) by a properly declared, ODK-style BFO import module.

This ensures that the BFO-owned classes distributed with RO are always kept in sync with their original BFO versions, as discussed in #107.


wdduncan:
Thanks @gouttegd !
I hold off any more work on [owns / is owned by](https://github.com/oborel/obo-relations/pull/705) until this is merged.

wdduncan:
@gouttegd I'm adding:
Fixes #107 

Delete this comment if you think this PR doesn't close the issue.

src/ontology/ro-odk.yaml
```diff
@@ -25,6 +25,8 @@ import_group:
         - dc:source
         - IAO:0000115
   products:
+    - id: bfo
+      make_base: TRUE
```
matentzn, line 29:
```suggestion
      module_type: custom
```

src/ontology/ro.Makefile
```diff
@@ -75,13 +75,6 @@ grepro:
 	cat ro-diff.txt | grep /RO_ > ro-diff.txt.ro.txt
 	cat ro-diff-core.txt | grep /RO_ > ro-diff-core.txt.ro.txt
```
matentzn, line 77:
```suggestion

# We are not using the BFO PURL here, just two manually curated BFO components.
imports/bfo_import.owl: components/bfo-axioms.owl components/bfo-axioms.owl
	$(ROBOT) merge -i components/bfo-axioms.owl -i components/bfo-axioms.owl -o $@
```

src/ontology/ro.Makefile
```diff
@@ -28,8 +28,8 @@ validate-using-oort: ro-edit.owl
 # Custom components
 # ========================================
 
-core.owl: components/core.owl components/bfo-axioms.owl components/bfo-classes-minimal.owl
-	$(ROBOT) merge -i components/core.owl -i components/bfo-axioms.owl -i components/bfo-classes-minimal.owl annotate --ontology-iri $(ONTBASE)/$@ $(ANNOTATE_ONTOLOGY_VERSION) -o $@
+core.owl: components/core.owl imports/bfo_import.owl
+	$(ROBOT) merge -i components/core.owl -i imports/bfo_import.owl annotate --ontology-iri $(ONTBASE)/$@ $(ANNOTATE_ONTOLOGY_VERSION) -o $@
```
matentzn, line 32: Lets not make any updates to core release better..

matentzn (changes requested):

src/ontology/ro.Makefile
gouttegd: Err, sorry, but what good does that make? This does not fix anything, _unless_ we also update once the `components/bfo-axioms.owl` to synchronize it with the current version of upstream BFO – in which case this PR becomes an implementation of “solution 2” [described here](https://github.com/oborel/obo-relations/issues/107#issuecomment-1485063040). This just turns the manually curated components into a manually curated pseudo-import.

gouttegd: I’d argue it’s even worse than the current situation, but this would make it _seem_ like we are importing upstream BFO, while in reality we are still using a manually curated frozen (and outdated!) version of it.

matentzn: Sorry I misunderstood something in slack - now I got it. If I understand it correctly and purl.../bfo.owl in its current form actually more or less corresponds to a merge of bfo-axioms with bfo-classes-minimal, we should go ahead with what you had. Sorry about the confusion!

gouttegd: > purl.../bfo.owl in its current form actually more or less corresponds to a merge of bfo-axioms with bfo-classes-minimal

Yes. Except that the version published by the PURL server has updated definitions, while our manually maintained copy (or not so maintained actually) still has definitions from several years ago – hence the incompatibility!

The whole point of this PR is to use whatever version of BFO is currently published by the PURL server, so that when third-party ontologies want to import both purl.obolibrary.org/obo/ro.owl and purl.obolibrary.org/obo/bfo.owl, they can do so without being bitten by the fact that RO comes with its own outdated definitions of BFO classes.

gouttegd:
Closing here as it was decided to fix using the simpler route in #709.

PR #708, opened 2023-04-03
Demonstrate annotating affiliation to contribution

cthoyt:
In this PR, I demonstrate the following:

1. Declaring an individual corresponding to my current top-level affiliation, Harvard University (https://ror.org/03vek6s52) via its Research Organization Registry (ROR) persistent identifier
2. Add an annotation on to the Annotation Assertion that I am the `dcterms:creator` of [RO:0018030 (chemical relationship)](http://purl.obolibrary.org/obo/RO_0018030) that says I did this under my affiliation with Harvard University. This uses the http://www.w3.org/ns/org#memberOf predicate, as suggested in https://github.com/information-artifact-ontology/ontology-metadata/issues/127#issuecomment-1498014390.

What's **not** in scope of this PR:
- making decisions about whether you can use top-level organizations as creators/contributors. For now, this is fair game
- annotating more granular things like synonyms, which requires higher order annotations (we ain't got time for this)

Screenshot 2023-05-15 at 17 25 41

## Why

Given we have begun standardizing contributions using canonical ORCID URIs and connecting the [ORCID Instance Ontology](https://github.com/cthoyt/orcidio) to make those annotations rich in metadata, the next step is to get more granular and include institutional affiliations as well.

In the past, some contributions have been annotated to entire organizations (e.g., GO Constortium), to meetings/workshops, etc. Working towards a well-defined model can allow more precise information that focuses on who exactly is doing the typing/editing, then still maintain the provenance on how they did it. Potentially multiple organizations can be annotated to a given `dc:creator` Annotation Assertion in OWL, corresponding to the affiliations of the contributor at the time of contribution, referencing the workshop/scenario in which they made the contribution, etc. One person might have made contributions over time within the context of their work in different organizations  (e.g., after moving institutions), and therefore may appear with different annotations in different records.

In addition to URIs from ROR, which itself is very picky about only including top-level organizations in its controlled vocabulary, Wikidata can be used for more granluar things like specific departments, workshops, etc.

## Benefits

1. Better reporting for organizations. If you're an organization that has many different people working on many different ontologies, you might want to be able to say how many terms you created/contributed to. The solution proposed in this PR is also slightly better than the two step mapping from organization -> person -> contribution since the same person might be part of different organizations over time and making contributions in different capacities/with different affiliations.
2. Better reporting for the community. After beginning to adopt this kind of annotation, it can be shown across the entire OBO Foundry community (and other ontologies that adopt OBO standards) what organizations are making contributions

## TO DO, before merging this PR

- [x] Pick (or mint) the right relation for this
- [ ]  Collect additional ideas to add to the "Benefits" list

## Follow-up, after this PR

1. Write documentation/tutorials on how to make these annotations
2. Present this annotation concept to the greater OBO community
3. Pull in the [ROR Instance Ontology (RORIO)](https://github.com/cthoyt/rorio) to auto-propogate the appropriate named individual declarations, similarly to how ORCIDIO is imported.


matentzn:
As you know, I fully endorse the sentiment to document organisations. I think RORIO is the way to go, and for those orgs that don't have ROR, we provide a standard wikidata module.

Your proposal as far as I can see has one big downside which is that a lot, if not most, attribution happens on level 2 in the axiom assertions stack:

```
 <owl:Axiom>
        <owl:annotatedSource rdf:resource="http://purl.obolibrary.org/obo/HP_0000002"/>
        <owl:annotatedProperty rdf:resource="http://purl.obolibrary.org/obo/IAO_0000115"/>
        <owl:annotatedTarget>Deviation from the norm of height with respect to that which is expected according to age and gender norms.</owl:annotatedTarget>
        <oboInOwl:hasDbXref rdf:resource="https://orcid.org/0000-0002-0736-9199"/>
    </owl:Axiom>
```

Adding another axiom annotation on that level will not fly, as it involves triple nesting, for which we have zero tool support.

Perhaps this is best solved with an SOP?

- Ask people to always add an ROR when they make and edit/add a term
- Import RORIO the same way we import ORCIDIO

cthoyt:
Just to be clear, you mean that the non-protege tooling doesn't support doing things with the annotations on Annotation Assertions (which, by the screenshot, you can see at least protege itself supports), and the alternative is to just use a simple `dc:contributor` (or equivalent relation) in the axiom itself?

The downside to this is it would rely on assuming the connection between the contributor and the organizations, but I guess this could be acceptable in order to simplify things.

matentzn:
I am not talking about annotations on annotation assertions, I am talking about axiom annotations on axiom annotations on annotation assertions (maybe this is what you mean, just to be clear). Your screen shot is only an axiom annotation on an annotation assertion. What I am talking about is having the same statement (dc:contributor X) on a synonym assertion, like:

a oio:hasExactSynonym { dc:contributor X }.

yes your downside is precisely the counterargument!

cthoyt:
ohh I see. So this current one would work but the synonym wouldn't since it's one level deeper

cthoyt:
In 8da4406, I demonstrate what it looks like to add this kind of axiom onto a synonym - there's no simple way to make some transitivity between Chris Mungall and LBNL in this example, unfortunately :/ 

Screenshot 2023-04-03 at 13 11 01

matentzn:
Looks nice. BTW, to make this particular suggestion fly, you not to know that _obo format_ cannot handle this very well. This is not important for RO, and @balhoff has implemented a fix in OWLAPI afaik, but just want to be complete here :)

cthoyt:
I actually would rather maintain the most granular annotations. What I would suggest is:

1. Only add contributor information for the person on synonyms (and explicitly not the affiliation-level information)
2. Make sure that all people who are listed as a contributor on a synonym is also listed as a contributor on the term itself w/ `dc:contributor`
3. Include the high-granular information on the top-level. If the same person makes multiple contributions over time with different affiliation, I dont think there's any harm in the top-level term's dc:contributor annotation assertion having the union of all of the affiliations for that person over time

matentzn:
I think I totally agree with this suggestion. I want to wrote a postprocessing query for ODK that basically just turns owl axiom annotation assertion contributors in top level contributors - its super easy, and we could all agree that for most purposes and analyses, we only actually care about the top level.

cthoyt:
@matentzn I re-wrote the header of this PR, I hope it's not controversial anymore now that I made the scope explicitly smaller and we can consider this for merge.

jhpoelen:
@cthoyt Hi! We are reviewing this pull request in today's RO meeting. https://docs.google.com/document/d/19dhzj5QoliQZQgDJRczI878rLtexzMDo1g0K9_LRUWk/edit . Are you still interested to have your pull request reviewed? If not, please close this pull request. If so, please comment.

cthoyt:
@jhpoelen Yes, I would still like this to be considered. The short version is we still would need to decide what's the right property to start adding these annotations

PR #672, opened 2023-01-18
added new object property has roost for issue #453

diatomsRcool:
Added new object property "has roost" as child of "has habitat" to address issue #453

PR #709, opened 2023-04-03
Fix BFO incompatibility

gouttegd:
This is another way of fixing #107, as an alternative to #707.

It keeps the manually curated copy of BFO (as components `bfo-classes-minimal.owl` and `bfo-axioms.owl`), but updates the definitions on some the classes to match the definitions used in the version of BFO that is published at `purl.obolibrary.org/obo/bfo.owl`, so as to avoid clashing definitions when importing RO and BFO in the same ontology. Original definitions in the `bio-classes-minimal.owl` component are turned into comments.

This is the least intrusive way of fixing #107.

The BFO components are also properly imported into the -edit file.


wdduncan:
@gouttegd Thanks for updating the definitions. I noticed that the editor notes and examples of usage are not being pulled into the component/import. E.g., see [material entity()https://ontobee.org/ontology/BFO?iri=http://purl.obolibrary.org/obo/BFO_0000040) on Ontobee.  
Having these annotations helps clarify the meaning, although, I suppose a case could be case for not including them.

However, I think it would be helpful include the `http://www.w3.org/2000/01/rdf-schema#isDefinedBy: http://purl.obolibrary.org/obo/bfo.owl` annotation. There has been discussion in OBO about how to manage IRIs that migrate to be managed by another ontology. For example [part of](http://purl.obolibrary.org/obo/BFO_0000050) has a `BFO` namespace, but (as far as I know) it is now managed by `RO`.

gouttegd:
The conflicts with the current master branch are easy to fix, but the master branch is currently broken since #672, so I’d rather wait until it is fixed (which should be done with #731) before fixing the conflicts here.

wdduncan:
There are many annotations mismatches that are breaking things. I don't know what the right way to fix all these are. I didn't implement the annotation qc checks that are causing issues (e.g., dc vs dcterms).

Issue #716, opened 2023-04-28
import COB into RO


wdduncan:
@cmungall @matentzn @nataled @balhoff @jamesaoverton:
On the 2023-04-25 call, @bpeters42 mentioned importing COB in order allow for assignment of domains and ranges. E.g., #712 mentions `measurement data item` in the definition. 

This sounds reasonable to me, but what are your thoughts.

wdduncan:
If we are going to import COB terms, we need to decide if the import will have COB IRI or the IRI of the equivalent 3rd party term (e.g., OBI:organism).  
At present, we can't use COB to import 3rd party terms.

matentzn:
> we need to decide if the import will have COB IRI or the IRI of the equivalent 3rd party term

Whatever cob.owl says! If the terms are officially rewired during the COB release process, they should be used, i.e. https://www.ebi.ac.uk/ols4/ontologies/cob

> At present, we can't use COB to import 3rd party terms.

Very interesting problem! This requires a COB issue "Who owns COB terms imported from 3rd parties?"  and a bit of discussion.. In my opinion, something super weird should be done: The original IDs should be effectively ceded to COB (I am only 50,01% sure this is a good idea though, and 55% that its going to be very difficult). Very interesting point, thanks for raising it.

PR #717, opened 2023-05-23
measures characteristic

ddooley:
and inverse 'characteristic measured by'


src/ontology/ro-edit.owl
bpeters42, line 1: - Diff has a bunch of lines moving around. More than expected. 
- Inverse property is incorrect,.

bpeters42 (changes requested):
Diff has a bunch of lines moving around. More than expected.
Inverse property is incorrect,.

src/ontology/ro-edit.owl
ddooley: Fixed Inverse of 'measures characteristic'
Added inverse of axiom.

wdduncan:
Since the the relation holds between a process and a quality, do we to make the label more reflective of that? E.g.:
- `process measures characteristic`
- `characteristic measured by process`

This would make it clear that it is not a datum that holds measurement information.

wdduncan:
Since the COB import has been added to the ODK file, I think this addresses #716.   
Do others agree?

ddooley:
Re "process measures characteristic" label etcl, I feel it is intuitive that "measures" is tied to a measuring process, so no need to spell that out in label.

wdduncan:
I agree 'measures' brings to mind a process, but I thought it may need to be made clearer what is doing the measuring. 
E.g., When I step on scale, there is a measurement that happens. But, just reading the label `measures characteristic`, one might not understand whether it is it the scale (i.e., device) that measures my weight or the applying of pressure (i.e., process) that does the measuring?  
For the `measures characteristic` relation, it is the latter, but I can see how someone might think it is the former (i.e., device).

bpeters42:
I added my comments to the issue rather than this pull request so that they can viewed in context.

ddooley:
I'm going to have to redo this one.  Its getting too complicated trying to resolve diffs with pre-UBERON integration

anitacaron:
Hi @ddooley, I'll update this PR to align with the master. You don't need to worry about redoing the work.

wdduncan (approved):

anitacaron:
Note: COB import is empty. If there isn't a COB term to import, I'd recommend removing the COB import.

ddooley:
Yes I guess that COB import can be removed. I'd put it in thinking I could import OBI terms that way, but to my surprise one can't fetch 3rd party terms from COB indirectly (as mentioned in last RO meeting).

anitacaron:
There's another issue to address the COB import #716. It will be added when needed.

wdduncan:
Can you leave the COB import? I need it for one of my PRs. This will keep me from having to recreate it.

anitacaron:
I'll add it back, fixing issue #716.

wdduncan:
Thanks @anitacaron

ddooley:
The range was meant to be COB characteristic, but for now that translates to PATO:00000001 quality.

As for domain, @bpeters42 I recall you suggested "assay" for domain since that was doing the measuring, and I recall assay can't measure nonmaterial things like rates directly, but our solution is that assay can measure material things that are proxies for things like rates of change.  So I've changed label to "assay" on that basis. Unfortunately it added some superfluous diffs too tho.

ddooley:
@anitacaron want me to just do a redo on this one?  Might be simpler.

anitacaron:
@ddooley you can change the range to BFO:0000020 now.

src/ontology/ro-edit.owl
ddooley: I think @anitacaron fixed most of the diffs by updating this branch with recent master.

ddooley:
Switch to BFO:0000020 is now done!

ddooley:
I think @bpeters42 conversation about big diff is also taken care of with your work.

ddooley:
I did another merge of master into this branch and I think outstanding issues are taken care of?

src/ontology/ro-edit.owl
```diff
@@ -5988,6 +5990,25 @@ AnnotationAssertion(rdfs:label obo:RO_0009005 "has primary substance added"@en)
 SubObjectPropertyOf(obo:RO_0009005 obo:RO_0009001)
 IrreflexiveObjectProperty(obo:RO_0009005)
 
+# Object Property: obo:RO_0009006 (assay measures characteristic)
+
+AnnotationAssertion(obo:IAO_0000115 obo:RO_0009006 "A relation between an assay and a characteristic, in which the assay generates a data item which is a measure of a characteristic."@en)
```
anitacaron, line 5995: Could you add an example of usage, please?

src/ontology/ro-edit.owl
```diff
@@ -5988,6 +5990,25 @@ AnnotationAssertion(rdfs:label obo:RO_0009005 "has primary substance added"@en)
...
+AnnotationAssertion(obo:IAO_0000117 obo:RO_0009006 <https://orcid.org/0000-0002-8844-9165>)
```
anitacaron, line 5996: Could you please change the annotation to `terms:contributor`?

src/ontology/ro-edit.owl
```diff
@@ -5988,6 +5990,25 @@ AnnotationAssertion(rdfs:label obo:RO_0009005 "has primary substance added"@en)
...
+AnnotationAssertion(terms:date obo:RO_0009006 "2023-05-23T15:24:15Z"^^xsd:dateTime)
+AnnotationAssertion(rdfs:label obo:RO_0009006 "assay measures characteristic"@en)
+InverseObjectProperties(obo:RO_0009006 obo:RO_0009007)
+ObjectPropertyDomain(obo:RO_0009006 obo:OBI_0000070)
+ObjectPropertyRange(obo:RO_0009006 obo:BFO_0000020)
+
+# Object Property: obo:RO_0009007 (characteristic measured by assay)
+
+AnnotationAssertion(obo:IAO_0000115 obo:RO_0009007 "Inverse of 'measures characteristic'"@en)
+AnnotationAssertion(obo:IAO_0000117 obo:RO_0009007 <https://orcid.org/0000-0002-8844-9165>)
```
anitacaron, line 6006: Could you please change the annotation to `terms:contributor`?

anitacaron (changes requested):

src/ontology/ro-edit.owl
ddooley: Done!

src/ontology/ro-edit.owl
ddooley: done!

ddooley:
@wdduncan @bpeters42 final touchups are finished!

anitacaron (approved):
Technically, I approve. It still needs an ontology review.

src/ontology/ro-edit.owl
anitacaron: Thanks!

wdduncan (approved):

src/ontology/ro-edit.owl
```diff
@@ -6002,6 +6004,26 @@ AnnotationAssertion(rdfs:label obo:RO_0009005 "has primary substance added"@en)
...
+AnnotationAssertion(terms:contributor obo:RO_0009006 <https://orcid.org/0000-0002-8844-9165>)
+AnnotationAssertion(terms:date obo:RO_0009006 "2023-05-23T15:24:15Z"^^xsd:dateTime)
+AnnotationAssertion(rdfs:label obo:RO_0009006 "assay measures characteristic"@en)
+InverseObjectProperties(obo:RO_0009006 obo:RO_0009007)
+ObjectPropertyDomain(obo:RO_0009006 obo:OBI_0000070)
+ObjectPropertyRange(obo:RO_0009006 obo:BFO_0000020)
+
+# Object Property: obo:RO_0009007 (characteristic measured by assay)
+
+AnnotationAssertion(obo:IAO_0000115 obo:RO_0009007 "Inverse of 'assayt measures characteristic'"@en)
```
bpeters42, line 6020: typo: assayt --> assay

bpeters42:
I think the pull request as is doesn't do much harm; there is still a lot to be resolved for when we transition to modeling values of measurements via @jamesaoverton approach, but this will not make that worse, and it has been requested for a while. So I would approve. (apart from the minor typo I found)

jamesaoverton:
@bpeters42 I have concerns about the bigger picture here, but I agree that this PR should go ahead.

src/ontology/ro-edit.owl
anitacaron: done

anitacaron (approved):
Approving again after fixing a typo in the inverse property's definition.

wdduncan (approved):

Issue #710, opened 2023-04-12
New term request: ‘positively correlated with’ and ‘negatively correlated with’


CDowland:
This request is for two subtypes of the preexisting RO term 'correlated with' (http://purl.obolibrary.org/obo/RO_0002610), which is annotated as follows:
```
Definition: A relationship that holds between two entities, where the entities exhibit a statistical dependence relationship. The entities may be statistical variables, or they may be other kinds of entities such as diseases, chemical entities or processes.

Curator note: Groups both positive and negative correlation
```

I’m requesting 'positively correlated with' and 'negatively correlated with' as subtypes of 'correlated with'.

## Preferred term label
`positively correlated with`

## Textual definition
`Correlation for which the statistical dependence is positive.`

## Suggested parent term
`RO: correlated with (http://purl.obolibrary.org/obo/RO_0002610)`

## Attribution
`ORCID ID: 0000-0003-1909-9269`

## Preferred term label
`negatively correlated with`

## Textual definition
`Correlation for which the statistical dependence is negative.`

## Suggested parent term
`RO: correlated with (http://purl.obolibrary.org/obo/RO_0002610)`

## Attribution
`ORCID ID: 0000-0003-1909-9269`

As the definition for 'correlated with' notes, a variety of ontological types can occupy the domain and range positions. No domain and range are specified for it, which seems like the right approach for these subtypes as well. 

The definitions are based on the definitions of 'correlated with,' adapted to indicate whether the correlation is positive or negative. But perhaps there are clearer/better ways to define them.

wdduncan:
The definition references the use of statistics, but the label does not. Not all "correlations" are statistical in nature. If you intend this relation to be about things calculated using statistical methods, then we suggest modifying the label to reflect this.

CDowland:
I'm proposing these as subtypes of RO: 'correlated with', which is indeed a statistical relation, and likewise mentions statistics in the definition but not in the label.

CDowland:
Below is a revised request for the two terms that I hope avoids the concerns raised.  

Label: positively correlated with
Alternative term: directly correlated with
Definition: A relation between entities in which one increases or decreases as the other does the same.
Definition source: [](https://dictionary.apa.org/positive-correlation)
Suggested parent term: RO: correlated with [](http://purl.obolibrary.org/obo/RO_0002610)
Characteristics: Symmetric, Reflexive
Attribution: ORCID ID: 0000-0003-1909-9269

Label: negatively correlated with
[...]

PR #719, opened 2023-05-25
New terms from issue 710

CDowland:
Adds terms 'positively correlated with' and 'negatively correlated with' from issue #710


wdduncan (changes requested):
Please change [created_by](http://www.geneontology.org/formats/oboInOwl#created_by) annotation to either [dcterms:creator](http://purl.org/dc/terms/creator) or [dcterms:contributor](http://purl.org/dc/terms/contributor). I prefer the latter, but the former is acceptable too.

There has been a lot of discussion within the Foundry about standardizing which annotation is used for term creation attribution. You may not be aware of these discussions. @matentzn or @cthoyt can weigh in if they wish to provide more background.

wdduncan (approved):

CDowland:
@wdduncan Thanks!

Issue #712, opened 2023-04-13
NTR: 'contains measured amount'


jmwhorton:
Note: this relationship is currently in OBIB (http://purl.obolibrary.org/obo/OBIB_0000038)

## Preferred term label

has actual load

## Textual definition

A relation that relates a container to a measurement datum that specifies the actual amount of material in the container.

## Definition Source

Penn Group

## Term Editor

Person: Jie Zheng, Chris Stoeckert

wdduncan:
By 'amount' do you mean its weight? Or volume? Or would categorical units be allows, e.g. 'half full'.

It was also suggested that 'actual' be removed from the name.

jmwhorton:
I'm going to tag @zhengj2007 and @cstoeckert to address any feedback on this term.

cstoeckert:
For the same reasons as commented in https://github.com/oborel/obo-relations/issues/711 I think this should apply to both weight and volume. I also think including "actual" in the label helps distinguish which relation to use for the case of a 50 ml conical tube (has max capacity) containing 25 ml (has actual load). The restriction is for a measurement datum so categorical units are fine for that in my opinion.

nataled:
Definitely not 'has actual capacity' since that would be synonymous with 'has maximum capacity'. And although I think it's mildly superfluous, I don't have much objection to including 'actual' in the name since people often refer to 'actual amounts' of things.

wdduncan:
> Definitely not 'has actual capacity' since that would be synonymous with 'has maximum capacity'.

Right! My mistake :)

> I don't have much objection to including 'actual' in the name since people often refer to 'actual amounts' of things.

Sure ... In most (maybe all) contexts I've dealt with dealing with measurements (e.g., weight, volume, height, duration), 'actual' was assumed, and max/min had to be stated.

wdduncan:
> How about 'has contained amount' ?  

'contains amount' sounds more natural to me, but I'm okay with the label 'has contained amount'.  

The way definition current reads it would be necessary to import `measurement datum` to formally specify the range. One way to do this would be to create a COB import that includes `measurement datum`. @matentzn do you suggest a better way?

@cstoeckert If I assign an IRI for the term are you (or someone in your group) able to make a PR for this?

cstoeckert:
@wdduncan yes I or someone from OBIB can work on this. Thanks.

wdduncan:
Thanks @cstoeckert   
Please use `RO_0017002` as the namespace id in the new term.

[...]

PR #726, opened 2023-06-12
NTR 'contains measured amount'

jmwhorton:
Addresses new term request: #712 (original label was 'has actual load')

Please note:  imports were added manually to other_import.owl.  I couldn't see another place that made more sense.  If I have added the imports incorrectly, please let me know and I'm happy to make the change.


wdduncan:
@jmwhorton please adress the following:

1. We can discuss at the next meeting, but I would prefer the label to more explicitly reflect that the range is a measurement data. E.g., "container amount measured by". "contains amount" can still be a synonym (e.g., [has_exact_synonym](http://www.geneontology.org/formats/oboInOwl#hasExactSynonym).

2. Change 
```
A relation that relates a container to a measurement datum that specifies the actual amount of material in the container.
```
to
```
A relation between a container and measurement datum that specifies the actual amount of material in the container.
```

3. Instead of `term editor` (IAO_0000117), use annotation `dcterms:contributor` and the editor's ORCID. Each contributor gets their own contribution annotation.  
As an example see [positively correlated with](http://purl.obolibrary.org/obo/RO_0017003). (You will have to open in Protege. OLS doesn't pull up the term.)

4. Can you please provide an [example of usage](http://purl.obolibrary.org/obo/IAO_0000112).

5. The [definition source](https://ontobee.org/ontology/IAO?iri=http://purl.obolibrary.org/obo/IAO_0000119) annotation doesn't seem relevant here. "Contains" is a common word and the contributors are listed if anyone has questions.

6. Is there an inverse relation? E.g., "measures container amount"

wdduncan:
@nataled I see your point.

> "this beaker 'container amount measured by' Mettler scale".

I made the suggestion b/c I was trying to make it more clear that the range is a `measurement datum` and not a literal.

nataled:
@wdduncan substituting 'as' in place of 'by' might come a bit closer (in my mind, at least) in evoking the correct range type. Even so, the 'as' version could just as likely have someone think that 'weight' or 'volume' would be an appropriate response. I'm not sure how best to do it.

jmwhorton:
I have addressed numbers 2, 3, and 5 from @wdduncan's list above.  For the rest, I'll defer to the group consensus and ask @cstoeckert for an opinion.

wdduncan:
@nataled I've thinking about this past few days, and I don't have any real answer regarding `container amount measured by` vs. `container amount measured as`. We can discuss at tomorrow's RO meeting. But here are some random thoughts.

1. It would be nice if the label `contains amount` had a nice inverse. Not sure what this is though ... Perhaps `amount in container`?
2. Normally, I'd be okay with using the label `contains amount`. But, there has been recent discussion about including data properties in RO. Maybe I'm worrying to much, and we should just use the label as is. If a data property for amounts in containers is proposed, we can address it then. (i.e., for now we kick the can down the road).

In any case, let's focus on wrapping this up at the meeting. This issue doesn't need to hang around too long.

src/ontology/ro-edit.owl
```diff
@@ -566,6 +566,7 @@ Declaration(ObjectProperty(obo:RO_0016001))
 Declaration(ObjectProperty(obo:RO_0016002))
 Declaration(ObjectProperty(obo:RO_0016004))
 Declaration(ObjectProperty(obo:RO_0017001))
+Declaration(ObjectProperty(obo:RO_0017002))
```
wdduncan, line 569: Please change IRI to `obo:RO_0017005`. I made a mistake when I assigned `obo:RO_0017002`.

jmwhorton: RO_0017002 has been corrected to RO_0017005

wdduncan:
Agreed on RO call to change label to `contains measured amount`.

src/ontology/imports/other_import.owl
```diff
@@ -51,6 +57,41 @@ SubClassOf(<http://purl.obolibrary.org/obo/OGMS_0000031> <http://purl.obolibrary
 AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/UPHENO_0001001> "phenotype"@en)
 SubClassOf(<http://purl.obolibrary.org/obo/UPHENO_0001001> <http://purl.obolibrary.org/obo/PATO_0000001>)
 
+# Class: <http://purl.obolibrary.org/obo/OBI_0000967> (container)
```
anitacaron, line 60: There's an OBI import now. Please add the CURIE ids to the `src/ontology/imports/obi_terms.txt` and run `sh run.sh make imports/obi_import.owl`. Let me know if you need help, or I can do it.

anitacaron (changes requested):

src/ontology/imports/other_import.owl
wdduncan: @jmwhorton have you had a chance to review @anitacaron requested change?

jmwhorton: Apologies, it's been a busy few weeks.  I should be able to take care of it this week.  Appreciate you checking in on it.

anitacaron: Please make sure to pull the changes in your local branch version. 😄

wdduncan:
Fixes #712

src/ontology/imports/other_import.owl
jmwhorton: The obi terms have been moved, and the `make` command has been run.

jmwhorton:
Added example of usage.  I believe all issues have been addressed, and this is ready to be reviewed again.

wdduncan:
Thanks @jmwhorton  
Did you manually add the domain ([device](http://purl.obolibrary.org/obo/OBI_0000967)) and range ([measurement data item](http://purl.obolibrary.org/obo/IAO_0000109)] to `obi_import.owl` manually?

If so, I believe those terms should be added to the `obi_terms.txt` file, and the import should be rebuilt.  Otherwise, the changes you made to `obi_import.owl` will get overwritten by running ODK.

@matentzn @anitacaron can you confirm this?  I can't quite recall the ODK command to rebuild and OBI import.  I think it is:`sh run.sh make refresh-obi`. However, is it better to update the whole repo, i.e., `sh run.sh make update_repo`?

wdduncan:
P.S. I just noticed @anitacaron comment. Not sure if it better to run `sh run.sh make refresh-obi` or `sh run.sh make imports/obi_import.owl`. I think they may do the same thing.

jmwhorton:
For the sake of clarity I'll answer both the previous commends, @wdduncan.  I didn't add the imports to obi_import.owl manually.  I did run `make imports/obi_import.owl` 

However, do note that I added terms into other_import.owl manually (when I originally made the change, all of the imports were added there.  I manually added them to the .owl file since other_terms.txt was empty).

Happy to make any changes if I need to.

wdduncan:
Thanks @jmwhorton !  
I believe the new terms should go in `obi_terms.txt`. Let's wait on @anitacaron or @matentzn to confirm.

jmwhorton:
To be specific, the new terms that I left in `other_terms.owl` are IAO terms, not OBI terms.  (so I didn't want to put them in the OBI file)

wdduncan:
Thanks. I didn't catch that. Ideally, we should create new IAO import, but measurement data item is in OBI ... What you think @anitacaron ?

src/ontology/imports/other_import.owl
anitacaron: Thank you, @jmwhorton!

anitacaron:
You're right, @wdduncan. We need to create an IAO import and add the IAO term there instead of adding it to the `other_import.owl` file.

wdduncan:
@anitacaron Can you modify the ODK file to create the IAO import? Also, is there documentation about the purpose of `other_import.owl`?

ddooley:
One question - can the IAO terms be imported by way of COB?  I recall there was going to be some change made to COB import to enable 3rd party terms imported that way?

wdduncan:
@anitacaron will be able to modify the ro-odk.yaml so that IAO is included as an import?  
Also, do you have any documentation on what `other_import.owl` is used for?

anitacaron:
@wdduncan I think the `other_imports.owl` was created to have all imports, but it was manually updated. There's a comment in the file.

> Annotation(rdfs:comment "Ad-hoc additional imports, manually copied. This should be split into several dynamically imported sub-modules.")

Ideally, we should remove this file. Now it only imports a few terms. It would be good to review, update, and create a proper import in another issue. For example, instead of [disease (OGMS)](http://purl.obolibrary.org/obo/OGMS_0000031), use [disease (MONDO)](http://purl.obolibrary.org/obo/MONDO_0000001).

> Can the IAO terms be imported by way of COB?

Now that we have the COB import, you can import the COB terms like OBI. Add the terms in the `src/ontology/imports/cob_terms.txt` file and run `sh run.sh make imports/cob_import.owl`
I found [measurement datum - COB:0000121](https://www.ebi.ac.uk/ols4/ontologies/cob/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FCOB_0000121?lang=en) and [data item - IAO:0000027](https://www.ebi.ac.uk/ols4/ontologies/cob/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FIAO_0000027?lang=en). But only the `measurement datum` is used in the new RO property, so I think you only need to add this term to the `cob_terms.txt` file.

src/ontology/imports/other_import.owl
```diff
@@ -51,6 +54,25 @@ SubClassOf(<http://purl.obolibrary.org/obo/OGMS_0000031> <http://purl.obolibrary
...
+SubClassOf(<http://purl.obolibrary.org/obo/IAO_0000027> <http://purl.obolibrary.org/obo/IAO_0000030>)
+
+# Class: <http://purl.obolibrary.org/obo/IAO_0000030> (information content entity)
+
+AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/IAO_0000030> "information content entity")
+
+# Class: <http://purl.obolibrary.org/obo/BFO_0000040> (material entity)
+
+AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/BFO_0000040> "material entity")
+
```
anitacaron, line 75: Please, remove this entry because it's already defined in the `core.owl` artefact.

src/ontology/imports/other_import.owl
```diff
@@ -51,6 +54,25 @@ SubClassOf(<http://purl.obolibrary.org/obo/OGMS_0000031> <http://purl.obolibrary
 AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/UPHENO_0001001> "phenotype"@en)
 SubClassOf(<http://purl.obolibrary.org/obo/UPHENO_0001001> <http://purl.obolibrary.org/obo/PATO_0000001>)
 
+# Class: <http://purl.obolibrary.org/obo/IAO_0000109> (measurement datum)
+
+AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/IAO_0000109> "measurement datum")
+AnnotationAssertion(<http://purl.obolibrary.org/obo/IAO_0000115> <http://purl.obolibrary.org/obo/IAO_0000109> "A measurement datum is an information content entity that is a recording of the output of a measurement such as produced by a device."@en)
+SubClassOf(<http://purl.obolibrary.org/obo/IAO_0000109> <http://purl.obolibrary.org/obo/IAO_0000027>)
```
anitacaron, line 61: Please, import the COB term [COB:0000121](https://www.ebi.ac.uk/ols4/ontologies/cob/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FCOB_0000121?lang=en) instead and remove this entry here.

src/ontology/imports/other_import.owl
```diff
@@ -51,6 +54,25 @@ SubClassOf(<http://purl.obolibrary.org/obo/OGMS_0000031> <http://purl.obolibrary
...
+
+# Class: <http://purl.obolibrary.org/obo/IAO_0000027> (data item)
+
+AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/IAO_0000027> "data item")
+SubClassOf(<http://purl.obolibrary.org/obo/IAO_0000027> <http://purl.obolibrary.org/obo/IAO_0000030>)
+
+# Class: <http://purl.obolibrary.org/obo/IAO_0000030> (information content entity)
+
+AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/IAO_0000030> "information content entity")
```
anitacaron, line 70: Please, remove these additional classes related to the term above.

anitacaron (changes requested):

src/ontology/imports/other_import.owl
jmwhorton: Removed

src/ontology/imports/other_import.owl
jmwhorton: Removed

jmwhorton:
Requested changes have been made, @wdduncan

jmwhorton:
@anitacaron 'dc' has been replaced with 'terms' now to fit with recent changes to master.

wdduncan:
@anitacaron I don't understand which changes you requested. Have they been addressed?

anitacaron:
Yes, they were addressed, and all conversations were resolved. I'm waiting for someone from the ontology to approve it, and I'll approve it after.

wdduncan (approved):

wdduncan:
> Yes, they were addressed, and all conversations were resolved. I'm waiting for someone from the ontology to approve it, and I'll approve it after.

Ok. I approved.

wdduncan:
Thanks @anitacaron !!!!

PR #728, opened 2023-06-21
2023-06-28 Release

anitacaron:

PR #729, opened 2023-06-22
Update chemical relation xrefs to skos:exactMatch

cthoyt:
Suggested in https://github.com/oborel/obo-relations/pull/728#discussion_r1238352410, this PR switches the chemical relationships I've recently added to use skos:exactMatch instead of oio:hasDbXref.

### DIFF
--- a/src/ontology/ro-edit.owl
+++ b/src/ontology/ro-edit.owl
@@ -689,6 +689,7 @@ Declaration(AnnotationProperty(oboInOwl:inSubset))
 Declaration(AnnotationProperty(oboInOwl:source))
 Declaration(AnnotationProperty(rdfs:seeAlso))
 Declaration(AnnotationProperty(skos:closeMatch))
+Declaration(AnnotationProperty(skos:exactMatch))
 ############################
 #   Annotation Properties
 ############################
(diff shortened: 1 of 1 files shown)


src/ontology/ro-edit.owl
```diff
@@ -1224,18 +1225,42 @@ AnnotationAssertion(rdfs:label subsets:ro-eco "eco subset")
...
+
+AnnotationAssertion(rdfs:label rdfs:seeAlso "see also")
+
+# Annotation Property: skos:closeMatch (close match)
+
+AnnotationAssertion(rdfs:label skos:closeMatch "close match")
+
+# Annotation Property: skos:exactMatch (exact match)
+
+AnnotationAssertion(rdfs:label skos:exactMatch "exact match")
```
matentzn, line 1262: This information (labels etc) must be imported from OMO rather than hard coded in the edit file!

matentzn: Just remove all these AnnotationAssertions again and the PR is fine.

matentzn (changes requested):

src/ontology/ro-edit.owl
cthoyt: done

matentzn (approved):
Gr8

matentzn:
@anitacaron probably better to merge this and rerun RO release

PR #736, opened 2023-07-10
Import COB

### LINKED ISSUE #716: import COB into RO

anitacaron:
Fixes #716


src/ontology/Makefile
```diff
@@ -388,6 +388,16 @@ $(IMPORTDIR)/obi_import.owl: $(MIRRORDIR)/obi.owl $(IMPORTDIR)/obi_terms_combine
 		query --update ../sparql/inject-subset-declaration.ru --update ../sparql/inject-synonymtype-declaration.ru --update ../sparql/postprocess-module.ru \
 		$(ANNOTATE_CONVERT_FILE); fi
 
+## Module for ontology: cob
+
+$(IMPORTDIR)/cob_import.owl: $(MIRRORDIR)/cob.owl $(IMPORTDIR)/cob_terms_combined.txt
+	if [ $(IMP) = true ]; then $(ROBOT) query -i $< --update ../sparql/preprocess-module.ru \
+    extract -T $(IMPORTDIR)/cob_terms_combined.txt --copy-ontology-annotations true --force true --method BOT \
+    remove --base-iri $(OBOBASE)/COB --axioms external --preserve-structure false --trim false \
```
matentzn, line 396: I think you will have to use "custom" module type here, because this row

```
    remove --base-iri $(OBOBASE)/COB --axioms external --preserve-structure false --trim false \
```

Is wrong for COB (and only for COB).

anitacaron: copied the same rule without this line to ro.Makefile

ddooley: So this means now we can import 3rd party ontology terms via COB?

matentzn (approved):
Yeah looks good.

src/ontology/Makefile
anitacaron: Yes.

anitacaron:
Now each PR needs two approvals. If you're happy with this, please approve it, and I'll merge it.

wdduncan (approved):

PR #738, opened 2023-07-17
core.owl edit: add process to domain to range of concretizes

### LINKED ISSUE #650: add process to domain to range of concretizes
In [BFO-2020](https://github.com/BFO-ontology/BFO-2020/blob/master/21838-2/owl/bfo-2020.owl) the domain of `concretizes at some time` is `process or 'specifically dependent continuant'`. The domain of RO's  `concretizes` should be updated to include `process` for sake of consistency.

wdduncan:
- add process (BFO_0000015) to domain of concretizes
- update definition of concretizes to include process
- add process (BFO_0000015) to range of is concretized as
- update definition of is concretized as to include process

Fixes #650 

I added process to domains and ranges in the `core.owl` file. Let me know if I was supposed to update `ro-edit.owl` instead.


src/ontology/components/core.owl
```diff
@@ -78,9 +85,9 @@
     
 
 
-    <!-- http://purl.org/dc/terms/license -->
+    <!-- http://purl.org/dc/elements/1.1/license -->
 
-    <owl:AnnotationProperty rdf:about="http://purl.org/dc/terms/license"/>
+    <owl:AnnotationProperty rdf:about="http://purl.org/dc/elements/1.1/license"/>
```
wdduncan, line 90: @anitacaron @matentzn Any ideas on why this changed? I didn't change it manually. I thought `dcterms:license` was the correct annotation now.

wdduncan: Thanks @anitacaron   
I think we should change this in `core.owl` (i.e., change annotation to `dcterms:license`). I can do the change as part of this PR, but I don't want to interfere with other work you may be doing.

shawntanzk:
> I added process to domains and ranges in the core.owl file. Let me know if I was supposed to update ro-edit.owl instead.

I might not be the right person to review this as I do not know where the components/core.owl derives from (I see that it is a custom component but has no makepoint in ro.makefile) but I do realise that those terms are not in ro-edit.owl (as in they are pulled in my importing component). I guess one can add those triples in edit as just dangling triples, but that would be a bit ugly in the edit owl file, so I would like you assume adding it to core is right, but again, cant figure out where core.owl came from.

PS is there a reason for all the other additions/changes in components/core.owl? (eg addition of all the BFO terms, and rdf:description to owl:class)

src/ontology/components/core.owl
anitacaron: Ok

wdduncan:
@anitacaron I still need two reviews to approve.

jhpoelen:
fyi @cmungall @shawntanzk please consider reviewing @wdduncan 's pull request related to "add process to domain to range of concretizes"

shawntanzk:
@jhpoelen sorry this slipped out of my radar when I changed job. As with my previous comment, I am not sure I am qualified enough to review this PR. Would need replies to my questions above before I can make a call. Thanks

jhpoelen:
@shawntanzk thanks for taking the time to respond and for sharing your notes.

@wdduncan can you please review the failed QC for this pull request? Also, please nominate other reviewers as hinted in @shawntanzk earlier comment.

wdduncan:
> @shawntanzk thanks for taking the time to respond and for sharing your notes.
> 
> @wdduncan can you please review the failed QC for this pull request? Also, please nominate other reviewers as hinted in @shawntanzk earlier comment.

@jhpoelen There is some issue going on with `components/core.owl`. When I tried to open the file locally, I got the error:
> org.xml.sax.SAXParseException; systemId: file:/Users/duncanw/repos/OBO/obo-relations/src/ontology/components/core.owl; lineNumber: 14; columnNumber: 45; Attribute "xmlns:terms" was already specified for element "rdf:RDF".

I copied the `core.owl` from the `master` branch, and now `core.owl` opens w/o any issues. `git diff` did not return any changes. So, I will push `core.owl`, although I don't know what changed.

shawntanzk:
Looks like theres no diff now lol - pulling this in would make no difference >.< 
@wdduncan  I think the changes you made were in core.owl so when you brought it over from master, it removed all the changes. I see that in the branch now, there isn't the domain and ranges that you added, and the definition is back to the old one. 
I have pushed a reverted version of core.owl - please do check that it is the changes you want. Thanks

wdduncan:
Thanks @shawntanzk !
There was duplicate `xmlns:terms="http://purl.org/dc/terms/"` in the rdfs/xml file for some reason. I've deleted it.

wdduncan:
@balhoff @ddooley @bpeters42 @cmungall Can one you review this? I believe we've agreed to do this. The PR needs to be approved.

cmungall (changes requested):
See comments here:

https://github.com/oborel/obo-relations/issues/650#issuecomment-2284356976

cmungall (approved):
I have re-reviewed this in light of the original issue. I see it is just making axioms logically consistent with bfo2020-core, which seems reasonable as it shares the label.

As an aside, I attempted to crystalize the policy here:

- #807

anitacaron (approved):
The `dc:license` annotation is just defined and not being used. I'll go through all the files and remove them in another PR.

PR #739, opened 2023-07-18
Obsoleting contains and contained in 693

### LINKED ISSUE #693: should contained_in be a sub-property of located_in?
I didn't see this question mentioned in earlier issues, nor do I see it addressed in the descriptions of the respective properties, so:

As best I can tell, `contained_in` ([RO:0001018](http://purl.obolibrary.org/obo/RO_0001018)) is a special case of `located_in` ([RO:0001025](http://purl.obolibrary.org/obo/RO_0001025)) - would it make sense to acknowledge that by making the former a sub-property of the latter?

I'm trying to decide how best to use these in an ontology I'm developing, and insight on this question would be helpful.

cmungall:
Fixes #693

- Obsoleting contains and contained in.
- adding seeAlso to issue


cmungall:
I think the MP one is a false positive (because it imports UBERON). Same of COLAO

I made PRs for UBERON and ENVO and an issue for AISM

anitacaron (approved):

Issue #750, opened 2023-08-16
Something in the build pipeline goes wrong if a certain prefix is defined for the DC terms namespace


balhoff:
See https://github.com/oborel/obo-relations/pull/749#discussion_r1296147714

PR #753, opened 2023-08-17
Rename terms namespace

anitacaron:
Somehow the new defined expansion for the `dc` namespace in `ro-edit.owl` was not getting into account when running the release pipeline. Instead, it was changing to its more common use, dce.

This also changes the remaining `dc` annotations to use the `terms` one in the components files.

See #750


matentzn (approved):
Lgtm

balhoff (approved):

Issue #646, opened 2022-11-04
Missing definitions for some ecological interactions


nleguillarme:
Definitions are missing for the following ecological interactions (the list is not exhaustive):
- [eats](http://purl.obolibrary.org/obo/RO_0002470)
- [is eaten by](http://purl.obolibrary.org/obo/RO_0002471)
- [provide nutrients for](http://purl.obolibrary.org/obo/RO_0002469)
- [acquires nutrients from](http://purl.obolibrary.org/obo/RO_0002457)
- [parasite of](http://purl.obolibrary.org/obo/RO_0002444)
- [parasitized by](http://purl.obolibrary.org/obo/RO_0002445)
- [has pathogen](http://purl.obolibrary.org/obo/RO_0002557)
- [pathogen of](http://purl.obolibrary.org/obo/RO_0002556)

nleguillarme:
We are working on trophic interactions as part of the [soil food web ontology](https://github.com/soilfoodwebontology/sfwo) project. Maybe we can discuss this with the people involved and come up with some suggestions?

diatomsRcool:
Yes, I can help with these

PR #762, opened 2023-10-19
adding eco relations definitions

diatomsRcool:
Adding definitions for 

- eats
- is eaten by
- provide nutrients for
- aquires nutrients from
- parasite of
- parasitized by
- has pathogen
- pathogen of
For issue #646


diatomsRcool:
I'm sure these need some word-smithing.....

src/ontology/ro-edit.owl
```diff
@@ -1,4 +1,5 @@
 Prefix(:=<http://purl.obolibrary.org/obo/ro.owl/>)
+Prefix(dc:=<http://purl.org/dc/terms/>)
```
ddooley, line 2: I think RO now has it that prefix "terms" => http://purl.org/dc/terms/

src/ontology/ro-edit.owl
```diff
@@ -4215,13 +4217,15 @@ SubObjectPropertyOf(obo:RO_0002468 obo:RO_0002465)
 
 # Object Property: obo:RO_0002469 (provides nutrients for)
 
+AnnotationAssertion(obo:IAO_0000115 obo:RO_0002469 "A feeding relationship where one organism provides nutrition for another organism without being directly consumed. Examples include but are not limited to producing an exudate that is consumed by other organisms or transformation of nutrients into more labile forms.")
```
ddooley, line 4220: Where inverse relations occur I recall RO is moving towards the pattern of just saying in defn. that "inverse of [inverse relation label]".

diatomsRcool:
Yes, Sierra, I think the "acquires nutrients from" would work.

src/ontology/ro-edit.owl
```diff
@@ -4827,6 +4833,8 @@ InverseObjectProperties(obo:RO_0002556 obo:RO_0002557)
 
 # Object Property: obo:RO_0002557 (has pathogen)
 
+AnnotationAssertion(obo:IAO_0000115 obo:RO_0002557 "A host interaction where the smaller of the two members of a symbiosis causes a disease in the larger member")
+AnnotationAssertion(obo:IAO_0000115 obo:RO_0002557 "A host interaction where the smaller of the two members of a symbiosis is causes a disease in the larger member")
```
anitacaron, line 4837: Duplicated definitions: need to remove one of them.

src/ontology/ro-edit.owl
```diff
@@ -676,14 +676,14 @@ Declaration(AnnotationProperty(obo:valid_for_go_gp2term))
...
-Declaration(AnnotationProperty(terms:source))
-Declaration(AnnotationProperty(terms:title))
+Declaration(AnnotationProperty(dc:conformsTo))
+Declaration(AnnotationProperty(dc:contributor))
+Declaration(AnnotationProperty(dc:creator))
+Declaration(AnnotationProperty(dc:date))
+Declaration(AnnotationProperty(dc:description))
+Declaration(AnnotationProperty(dc:license))
+Declaration(AnnotationProperty(dc:source))
+Declaration(AnnotationProperty(dc:title))
```
anitacaron, line 686: Could you update the branch from master, please?

anitacaron (changes requested):

src/ontology/ro-edit.owl
anitacaron: Or maybe change the prefix in your Protégé?

ddooley:
The FoodOn team of course is interested in some of these term definitions, namely "eats" and "is eaten by".  For eats, "A feeding relationship where one organism consumes another (either in whole or in part) through a type of mouth or other oral opening." We were just wondering if there is a more limiting word or label that conveys semantics of feeding on a single organism (or piece) directly.  More generally people "eat" salt, water, and multi-component foods, but this isn't included in the trophic sense of "eats". However we haven't been able to think of a simple solution to distinguish eating broadly vs the specific animal or part case.

ddooley:
There's a similar comment for the "aquires nutrients from" shortcut relation.  I see cattle feed is often supplemented with salt.  "Animal feed salt is an essential addition to animal feed. Salt maintains the mineral balance of the animals at the right level and keeps your livestock healthy. It is advisable to opt for sea salt because this type of salt also contains additional essential minerals and trace elements."  This relation as proposed requires us to use some other relation to connect an organism and other material entity nutritional sources.

diatomsRcool:
I struggled with similar issues as well. Unfortunately, interactions are separated as abiotic/biotic and biotic/biotic higher up in the hierarchy. We would need two "eats" one for biotic/biotic and one for abiotic/biotic - but this isn't that good as many foods are a combination. The line between abiotic/biotic isn't always very sharp. What if we elevate the "eats" and "acquires nutrients from" relations to be children of biotically interacts with? I can change the definition to include abiotic foods.

ddooley:
That sounds like a good solution!  (And I can see that these relations can operate in parallel with the other ones like "kills" and "preys on").

diatomsRcool:
Do we think "eats" should be a child of "acquires nutrients from"? One possible reason we cannot do that is organisms that eat rocks for the purpose of helping with digestion.

ddooley:
Indeed, eating isn't always motivated by nutritional intake.  Rock is a good eample.  Also, arguably caffeine isn't doing anything for the consumer, but its a motivator for eating.

src/ontology/ro-edit.owl
anitacaron: @diatomsRcool, do you need help with fixing it?

diatomsRcool:
definitions updated

wdduncan:
> Do we think "eats" should be a child of "acquires nutrients from"? One possible reason we cannot do that is organisms that eat rocks for the purpose of helping with digestion.

Should `acquires nutrients from` be a child of `eats`? The defn for `eats` mentions oral openings. Doesn't seem like this would apply to plants, but plans to `acquires nutrients from` some source.

ddooley:
As well, covering all cases, Parenteral / intravenous feeding enables nutrients to be collected directly, so 'acquires nutrients from' is not a subclass of eats on that count either.

Is eats meant to be restricted to animal organisms?  The only other example I can think of are coniverous plants - e.g. venus fly traps.  But there digestive processes happen at surface of plant?

anitacaron (approved):
Technically, it's approved, but it still needs an ontology review.

cmungall (approved):

PR #764, opened 2023-11-03
Update README.md

matentzn:


anitacaron (changes requested):
sss

cmungall:
???

nataled:
Well, while you're at it, probably should remove " and wider OBO library" from the text.

PR #776, opened 2023-12-15
Rename SWRL variables to use valid URN format.

balhoff:
This change updates the URN format used by SWRL variables to match the URN spec. See this old OWL API issue: https://github.com/owlcs/owlapi/issues/732 

I assume these variables were created sometime before the OWL API fix. Fixing this will quiet some complaints from RDF validators like this:

```
09:30:18 WARN  riot            :: [line: 16707, col: 76] {W107} Bad URI: <urn:swrl#y> Code: 61/SCHEME_PATTERN_MATCH_FAILED in PATH: The scheme specific syntax rules are violated.
09:30:18 WARN  riot            :: [line: 16722, col: 68] {W107} Bad URI: <urn:swrl#x> Code: 61/SCHEME_PATTERN_MATCH_FAILED in PATH: The scheme specific syntax rules are violated.
09:30:18 WARN  riot            :: [line: 16723, col: 68] {W107} Bad URI: <urn:swrl#z> Code: 61/SCHEME_PATTERN_MATCH_FAILED in PATH: The scheme specific syntax rules are violated.
```


matentzn (approved):
It looks good, but I dont really know much about the wider context. Can you give me a sense of wether there are any risks involved in this change?

balhoff:
A risk might be that I used the wrong updated format for the IRIs, and then Protege wouldn't show you the variable name properly when you try to edit the rules. But I looked at the new version in Protege and the rules look the same.

matentzn:
Alright, thanks. I leave it in @anitacaron hands to push this forward :)

anitacaron (approved):

Issue #713, opened 2023-04-13
NTR: 'is contact information for'


jmwhorton:
Note: this relationship is currently in OBIB (http://purl.obolibrary.org/obo/OBIB_0000734)

## Preferred term label

is contact information about

## Textual definition

a is contact information for b, if a is a information content entity, b is a human being or an organization and a identifies a physical location or electronic resource that allows to initiate communication with a.

## Term Editor

Mathias Brochhausen

## Suggested parent term

is about (http://purl.obolibrary.org/obo/IAO_0000136)

wdduncan:
Should the name be 'is contact information for' or 'is contact information about'? The former seems to make more sense.

Can someone from OBIB please join the next call to discuss?

wdduncan:
Any updates on this?

jmwhorton:
Per 6-22 OBIB call, 'is contact information for' makes more sense.

If there aren't any other comments, @jmwhorton will make the pull request.

jmwhorton:
Hello!  Pull request is still incoming, should have some time this week.

wdduncan:
@jmwhorton Please use RO_0017007.

anitacaron:
This seems like an annotation property and should be added to the [OBO Metadata Ontology](https://github.com/information-artifact-ontology/ontology-metadata) (OMO) instead.

mbrochhausen:
Can you please clarify. My understanding was all relations go into the relation ontology. Did that change? thx

PR #778, opened 2023-12-29
NTR: 'is contact information for'

jmwhorton:
Addresses issue #713


src/ontology/imports/obi_import.owl
```diff
@@ -43,6 +42,47 @@ AnnotationAssertion(<http://purl.obolibrary.org/obo/IAO_0000117> <http://purl.ob
...
+organization itself, as long as we can subclass it (University, Corporation,
+Government Agency, Hospital), instantiate it (Affymetrix, NCBI, NIH, ISO,
+W3C, University of Oklahoma), and have it play roles.
+
+This leads to my proposal: We define organization through the statements 1 -
+3 above, but without an 'is a' statement for now. We can leave it in its
+current place in the is_a hierarchy (material entity) or move it up to
+'continuant'. We leave further clarifications to BFO, and close this issue
+for now."@en)
+AnnotationAssertion(<http://purl.obolibrary.org/obo/IAO_0000117> <http://purl.obolibrary.org/obo/OBI_0000245> "PERSON: Alan Ruttenberg")
```
cthoyt, line 78: please use appropriate ORCID ids for referencing people

src/ontology/ro-edit.owl
```diff
@@ -6571,6 +6572,18 @@ AnnotationAssertion(rdfs:label obo:RO_0017005 "contains measured amount"@en)
 ObjectPropertyDomain(obo:RO_0017005 obo:OBI_0000967)
 ObjectPropertyRange(obo:RO_0017005 obo:COB_0000121)
 
+# Object Property: obo:RO_0017007 (is contact information for)
+
+AsymmetricObjectProperty(obo:RO_0017007)
+IrreflexiveObjectProperty(obo:RO_0017007)
+AnnotationAssertion(obo:IAO_0000115 obo:RO_0017007 "a is contact information for b, if a is a information content entity, b is a human being or an organization and a identifies a physical location or electronic resource that allows to initiate communication with a.")
+AnnotationAssertion(terms:contributor obo:RO_0017007 <https://orcid.org/0000-0003-1834-3856>)
+AnnotationAssertion(obo:IAO_0000112 obo:RO_0017007 "555-555-5555 is contact information for John Smith.")
```
cthoyt, line 6581: does this make sense? we typically don't see string literals appearing as the subject of our triples. What about the relation in the foaf vocabulary for phone number http://xmlns.com/foaf/spec/#term_phone? Or http://xmlns.com/foaf/spec/#term_mbox for email?

mbrochhausen: Hi Charles, let me make sure I understand your point correctly: the example of use should specify "John Smith's phone number" is contact information for John Smith. And then the actual string literal would be connected as a value specification? Thanks for bringing this up.

jmwhorton: Updated the example of use to reflect Mathias's comment.

wdduncan (dismissed):

wdduncan:
`is contact information` is a subproperty of [is about](http://purl.obolibrary.org/obo/IAO_0000136). However, `is about` is not included in RO. 

There was [consensus](https://github.com/oborel/obo-relations/issues/653#issuecomment-1677610127) in #653 to add `is about`, but to give it an RO id. @mbrochhausen would a different cause issues for you? I am concerned about having a different IRI for such a widely used relation.

cc @cmungall @balhoff @bpeters42 @jamesaoverton @nataled @addiehl @matentzn

mbrochhausen:
Hi,

can someone in the RO community please let me know, what I can or need to do to move this forward. Thx

jmwhorton:
What needs to be done on this PR? The last update was from @mbrochhausen in June, asking what is needed to move this forward.

@wdduncan or @anitacaron Could you please let us know what we can do?  

Thank you.

wdduncan:
I believe got stalled over whether to make it a subproperty if `is about` ([above](https://github.com/oborel/obo-relations/pull/778#issuecomment-2020997703)). To move forward, I suggest we add the term without adding `is about` to RO.

jmwhorton:
@wdduncan Please correct me if I'm mistaken.  That would mean we don't have to change the pull request, correct?  (Apologies, I haven't looked at this in quite a while.)

wdduncan:
@jmwhorton I don't know ... I haven't looked at the PR in a long time too :)

jmwhorton:
That is fair!  The relevant detail is that  the new term is listed as a sub object property of IAO 'is about'.  Can that be left in?

SubObjectPropertyOf(obo:RO_0017007 obo:IAO_0000136)

cmungall:
I am against striping, this creates import issues. The whole hierarchy
should go in one ontology or the other

mbrochhausen:
I see Chris' point. I agree it might be better to submit it to IAO. We will happily do so, unless this generates an issue with OBO Foundry Principle 7 downstream. The way I read that principle, if a relation, such as 'is about' does not exist in RO, we can reuse that relations and its subproperties without being in violation of OBO Foundry principles. Is that correct?

jmwhorton:
Closing pull request.  Term will possibly be submitted to IAO.

PR #798, opened 2024-04-05
Remove logical axioms from COB import related to RO properties

### LINKED ISSUE #797: Remove redundant domain in `occurs in` (BFO:0000066)
Yes, but the result is that the actual domain is the intersection of the two. In this case BFO:0000015 is a subclass of BFO:0000003, so the actual domain is just BFO:0000015.

_Originally posted by @balhoff in https://github.com/oborel/obo-relations/pull/795#discussion_r1551958498_

anitacaron:
Fixes #797

### DIFF
--- a/src/ontology/ro.Makefile
+++ b/src/ontology/ro.Makefile
@@ -53,7 +53,7 @@ $(IMPORTDIR)/cob_import.owl: $(MIRRORDIR)/cob.owl $(IMPORTDIR)/cob_terms_combine
 	if [ $(IMP) = true ]; then $(ROBOT) query -i $< --update ../sparql/preprocess-module.ru \
     extract -T $(IMPORTDIR)/cob_terms_combined.txt --copy-ontology-annotations true --force true --method BOT \
 	remove $(patsubst %, --term %, $(ANNOTATION_PROPERTIES)) -T $(IMPORTDIR)/cob_terms_combined.txt --select complement \
-	remove --select "RO:* BFO:0000050* BFO:0000051* BFO:0000060* BFO:0000066*" --axioms "annotation" \
+	remove --select "RO:* BFO:0000050* BFO:0000051* BFO:0000060* BFO:0000066*" --axioms logical \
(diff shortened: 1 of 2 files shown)


src/ontology/ro.Makefile
```diff
@@ -53,7 +53,7 @@ $(IMPORTDIR)/cob_import.owl: $(MIRRORDIR)/cob.owl $(IMPORTDIR)/cob_terms_combine
 	if [ $(IMP) = true ]; then $(ROBOT) query -i $< --update ../sparql/preprocess-module.ru \
     extract -T $(IMPORTDIR)/cob_terms_combined.txt --copy-ontology-annotations true --force true --method BOT \
 	remove $(patsubst %, --term %, $(ANNOTATION_PROPERTIES)) -T $(IMPORTDIR)/cob_terms_combined.txt --select complement \
-	remove --select "RO:* BFO:0000050* BFO:0000051* BFO:0000060* BFO:0000066*" --axioms logical \
+	remove --select "RO:* BFO:0000050* BFO:0000051* BFO:0000060* BFO:0000066*" --axioms "annotation logical" \
```
matentzn, line 56: I guess it is ok in this particular case, but note that my approach with --base-iri --select internal is safer as you do not delete _extensions_ (imagine a subclass axiom in the importing ontology, say :A subClassOf RO:001. This axiom would not be deleted with my approach, but with yours, which is mostly not right.

However here, I think we can make an exception as we do not expect COB to extend RO properties.

matentzn (approved):

src/ontology/ro.Makefile
anitacaron: Yes, I know. I tried the other one, and it didn't remove the logical axioms we needed to remove.

matentzn: I know why: https://github.com/ontodev/robot/pull/1187

Its ok lets roll with your solution now! Thanks!

PR #800, opened 2024-04-21
Grammar fix in src/ontology/reports/ro.owl-obo-report.tsv

Daniel-Mietchen:
That typo seems to exist [in multiple places](https://github.com/search?q=%22This+properties+were+created+originally%22&type=code), so not sure here is the best place to fix it, but we gotta start somewhere.


matentzn (changes requested):
@Daniel-Mietchen thank you for the drive-by! 

The correct place to permanently fix this issue, i.e. the source of truth, is https://github.com/oborel/obo-relations/blob/master/src/ontology/ro-edit.owl#L936

PR #816, opened 2024-10-18
Add regulates to relations which propagate 'in taxon'.

balhoff:
The historical usage of 'regulates' assumes it is within a single organism. For this reason we have the term [regulates in another organism](http://purl.obolibrary.org/obo/RO_0002010) without that assumption. This PR adds 'regulates' to the relations which propagate 'in taxon'.

This is good for GO, but does seem to assume a biological context. I'm not sure if the definition of 'regulates' should be tweaked to make that context more clear.


cmungall:
We need to make sure this won't break existing ontologies or their downstream ontologies.

All regulates axioms are likely in GO - if GO passes with this then we should be good to GO. If GO axioms need changed then we need time for releases to percolate, otherwise mixing certain versions of RO and GO will introduce incoherencies.

pgaudet:
@cmungall 

Can we give some review period for concerned groups to evalute the impact on their ontologies? 

Otherwise we should look for an alternative strategy to propagate these TCs in GO.

balhoff:
I can try adding this to a test Ubergraph build to see if there are negative consequences.

ddooley:
Nov 4, 2024. Jim will do the test soon.

balhoff:
I checked this group of merged ontologies with this property chain, and didn't find any unsatisfiabilities: https://github.com/INCATools/ubergraph/blob/master/ci/test-merged.ofn

cmungall (approved):

anitacaron (approved):

Issue #787, opened 2024-02-12
Infer has_soma_location from part_of for neuron


dosumis:
See https://github.com/obophenotype/cell-ontology/issues/2246#issuecomment-1859409950 for background and discussion.

is_neuron o 'part of' o 'is_anatomical_entity' -> 'has soma location'

Use local reflexivity pattern to define is_neuron & is_anatomical_entity:

ObjectProperty: is_neuron
   Domain: neuron

  Class: neuron
   SubClassOf: is_neuron Self

dosumis:
CC @cmungall @balhoff

PR #817, opened 2024-10-29
Addresses #787 Infer has_soma_location from part_of for neuron

aleixpuigb:
Addresses #787 Infer has_soma_location from part_of for neuron
@dosumis


dosumis:
CC @balhoff

dosumis:
@aleixpuigb  - if you run the reasoner on your RO edits you'll find many inconsistent classes.

is_reflexive needs to be avoided in OWL as it is non-local (everything class in the ontology gets related to itself via the reflexive relation).  Your inconsistent classes are cases where your relation has => inferred classification that violates a disjointness axiom.

Instead we can use OWL 'Self' to create local reflexivity like this:

image

Having said that, I'm failing to get classification to work as expected.  

I made this probe class (to be deleted)

image

with the aim that it would be classified under 

image


image


Other axiomatization:

image
image
image


@balhoff  - am I missing something or is this outside of ELK 5 functionality?

dosumis:
Just realised that I should be using equivalentTo  for the Self axioms. Still not working with Elk though.

aleixpuigb:
is_reflexive needs to be avoided in OWL as it is non-local (everything class in the ontology gets related to itself via the reflexive relation). Your inconsistent classes are cases where your relation has => inferred classification that violates a disjointness axiom.

Thank you for the clarification, I was unfamiliar with local reflexivity.

> Instead we can use OWL 'Self' to create local reflexivity like this:

I thought this should be added in CL, is that correct?

I will push this commit for now.

balhoff:
@dosumis you will need to use whelk to get the Self inferences. It is in ROBOT, and here is a Protege plugin: https://github.com/INCATools/whelk/releases/tag/v1.1.1

The plugin is not quite up-to-date; I need to make a new build.

aleixpuigb:
Thank you very much, @balhoff. I have tried using whelk an it doesn't get the classification. I have confirmed that whelk is working, but not for this case.
image

dosumis:
Works for me: 
image

dosumis:
Note use of Equivalence axioms with Self restrictions.

dosumis:
@balhoff  - any reason we shouldn't switch the CL pipeline to use Whelk`?

balhoff:
> @balhoff - any reason we shouldn't switch the CL pipeline to use Whelk`?

It should be fine. Whelk is just a little slower for pipeline use.

dosumis (approved):

dosumis:
Hi @balhoff  - could you review too?  (2 approvals needed for merge). Thanks!

balhoff (approved):

aleixpuigb:
Thank you @dosumis and @balhoff for your help!

PR #826, opened 2025-01-27
2025-01-27 Release

anitacaron:


src/ontology/imports/cob_import.owl
```diff
@@ -53,18 +59,25 @@ Declaration(AnnotationProperty(<http://www.w3.org/2004/02/skos/core#exactMatch>)
...
-# Object Property: <http://purl.obolibrary.org/obo/BFO_0000054> (realized in)
-
-AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/BFO_0000054> "realized in"@en)
+# Object Property: <http://purl.obolibrary.org/obo/BFO_0000055> (realizes)
 
-# Object Property: <http://purl.obolibrary.org/obo/BFO_0000067> (contains process)
-
-AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/BFO_0000067> "contains process"@en)
+AnnotationAssertion(<http://purl.obolibrary.org/obo/IAO_0000115> <http://purl.obolibrary.org/obo/BFO_0000055> "Paraphrase of elucidation: a relation between a process and a realizable entity, where there is some material entity that is bearer of the realizable entity and participates in the process, and the realizable entity comes to be realized in the course of the process")
+AnnotationAssertion(rdfs:label <http://purl.obolibrary.org/obo/BFO_0000055> "realizes")
```
anitacaron, line 78: Reason QC failed:

ERROR	multiple_labels	realizes	rdfs:label	realizes
ERROR	multiple_labels	realizes	rdfs:label	realizes@en

COB import updated to have BFO:0000055 and moved away from defining the labels using the language. However, BFO:0000055 label in the `core.owl` file uses language.
Should we remove the language tag on the `core.owl` file?

matentzn: @jamesaoverton

jamesaoverton: I think the right thing to do is make COB use `"realizes"@en`. I started using a template for COB terms, which makes it annoying to change just one term to use a language tag, and if they all us `@en` then that could cause similar multiple_labels problems... I'll try a few things.

jamesaoverton: I tried a few things, and I have a few options, but I'm not sure which to choose. I'll get back to you tomorrow.

jamesaoverton: I think I want to fix this by using `@en` for all labels in COB. I created this issue for discussion: https://github.com/OBOFoundry/COB/issues/293

anitacaron: I suggested the change on RO so this wouldn't block the RO release, which has already been a while since there is the need for a new one.

jamesaoverton: I thought it would be a quick fix in COB, but my proposal got blocked, and then it took me weeks to build out the alternative. Now I have https://github.com/OBOFoundry/COB/pull/295 which should fix this problem for RO.

You do what you have to do.

anitacaron:
To have a clean PR and branch with a new release, I'm closing this PR.

PR #834, opened 2025-03-04
Remove unnecessary 'ro-eco' subset and update related configurations

### LINKED ISSUE #833: make: *** No rule to make target 'build/subsets/ro-eco.owl', needed by 'subsets/ro-eco.owl'.  Stop.
When running the RO release, this error occurs when generating the `ro-eco` subset.

It's defined in the custom Makefile as:

https://github.com/oborel/obo-relations/blob/f06235f868c3f5fb7bd8762bdf349e6f2b4adaa5/src/ontology/ro.Makefile#L130-L131

However, I don't see `build/subset` folder anywhere in the repository. It might get lost when moving to ODK or be a hidden folder. This subset has not been updated since 2 years ago.

If we comment out this rule in the custom makefile, then it will fall under the default subset workflow as:

https://github.com/oborel/obo-relations/blob/f06235f868c3f5fb7bd8762bdf349e6f2b4adaa5/src/ontology/Makefile#L617-L620

Any ideas? @balhoff @cmungall @jamesaoverton @matentzn

anitacaron:
Fixes #833


matentzn (approved):
Looks great!

jamesaoverton (approved):

matentzn (approved):

PR #844, opened 2025-05-13
NTR 'characteristic value of' and inverse 'has characteristic value'

ddooley:
To support qualitative and quantitative data modelling: https://qqv.ontodev.com/primer.html#changes-to-assay-modelling .


wdduncan:
I think the label should be more specific to indicate if the this an object property or a data property. The label "is value of" is not clear on this.

I thought one of the reasons for proposing [Qualitative and Quantitative Values](https://qqv.ontodev.com/primer.html#changes-to-assay-modelling) was to create a simpler design pattern for relating literal values to characteristics/entities. If the range of the proposed `is value of` term is an instances of a class, then it not clear (to me) how this can provide a better pattern for literals.

ddooley:
I've updated the labels to be "has characteristic value" and "is characteristic value of"!  Ready for review.

ddooley:
I guess one outstanding question is whether "has characteristic value" should be a sub-property of "has characteristic"?  I imagine so, so have made that change in pr.  One other question - all the inverses such as "function of" don't have the word "is" in front of them, so to follow that pattern we'd just label inverse as "characteristic value of"?  I've made that change.

ddooley:
A recent point is that the community needs working examples of these relations in order to get comfortable with them.  I will finish a more in-depth recipe model example with calculations based on a knowledge graph containing these relations.

PR #849, opened 2025-05-19
Addition of biomarker-related object properties.

nataled:


balhoff (changes requested):
Check the values of IAO:0000117; the ORCID should probably an IRI instead of a string. Also, check if 'created_by' is preferred by RO vs. dcterms:contributor (I'm not sure what our standard is).

nataled:
Fixed the IAO:0000117 object. Was a string literal, now an IRI.

nataled:
@balhoff I made the requested changes to IAO:0000117 object, changing string literal to IRI. Also, looking at the definition of IAO0000117, I see that 'created_by' is basically an undefined relation, but its plain meaning is redundant with the IAO term. I've thus removed the created_by uses.

balhoff:
@nataled I'm not totally sure how to fix. Maybe make a clean branch off of master, and just paste in the section involving your edits, then resave?

PR #853, opened 2025-05-27
Normalise rohom and temporal intervals

matentzn:
FYI @anitacaron this is just from saving the ontology with Protege 5.6 - merging into the #852 branch


anitacaron (approved):
Thanks for this :)

PR #856, opened 2025-06-16
symptom of

### LINKED ISSUE #846: 'symptom of'
We intend to use 'Has symptom' (RO_0002452) for the Oral Health and Disease ontology. However, it would also help to have an inverse 'symptom of' relation, which does not currently exist in RO. For the time-being, we created our own version of this relation in OHD, but this relation seems generally applicable enough that it should exist in RO. Can a 'symptom of' relation be added to RO? We'd replace ours with one from RO if this was added.

Finn1928:
-added inverse for 'has symptom'
-fixes #846


wdduncan:
@anitacaron I am involved in the creation of this term (Finn works for me). 
The PR is not complicated, but are you sure you want me to review my own work?

wdduncan (approved):

PR #857, opened 2025-06-24
2025-06-24 Release

anitacaron:


anitacaron:
I think it's because of the presence of the BFO:0000051 that is part of the terms to extract.

wdduncan (approved):

PR #865, opened 2025-09-02
Add new term 'has substitutable entity'

### LINKED ISSUE #863: NTR 'has substitutable entity'
Definition: Relates one individual to another which is capable of standing in for the first individual in its other relations to other individuals.

GO would like to use this relation in causal activity modeling. We sometimes have a generic entity in a model which plays specific roles in interactions with other entities in a cellular pathway, where that entity could be any of a number of specific gene products (for example).

cc @cmungall

balhoff:
Fixes #863.

Label: has substitutable entity

Definition: Relates one individual to another which is capable of standing in for the first individual in its other relations to other individuals.

Example of usage: Given individuals f, i, x, and y, where f is a 'molecular function', i is an 'information biomacromolecule', and x and y are instances of different types of gene products, if f 'enabled by' i within an OWL model of a biological pathway, and i 'has substitutable entity' both x and y, then either x or y are valid enablers of function f within the modeled biological context.

### DIFF
--- a/src/ontology/ro-edit.owl
+++ b/src/ontology/ro-edit.owl
@@ -6969,7 +6969,7 @@ SubObjectPropertyOf(obo:RO_0019002 obo:RO_0019000)
 
 # Object Property: obo:RO_0019003 (has substitutable entity)
 
-AnnotationAssertion(obo:IAO_0000112 obo:RO_0019003 "Given a 'molecular function' f, f is 'enabled by' an 'information biomacromolecule' i. i 'has substitutable entity' both x and y, instances of two different gene products, either of which may play the role of i in its biological context.")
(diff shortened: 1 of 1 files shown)


balhoff:
I will try to reword to clarify. All of the lowercase letters are meant to be individuals, and `i 'has substitutable entity' x` and `i 'has substitutable entity' y`. The idea is that there is an instance graph involving a placeholder individual `i`, where in practice either `x` or `y`, doesn't matter which, takes the place of `i`. I'm sure my example can be clearer.

balhoff:
@nataled I updated the example of usage to the following:

>Given individuals f, i, x, and y, where f is a 'molecular function', i is an 'information biomacromolecule', and x and y are instances of different types of gene products, if f 'enabled by' i within an OWL model of a biological pathway, and i 'has substitutable entity' both x and y, then either x or y are valid enablers of function f within the modeled biological context.

wdduncan (approved):

PR #866, opened 2025-09-05
Issue 860: indicates as synonym for is marker for

### LINKED ISSUE #860: add 'indicates' as synonym for 'is marker for'
Add 'indicates' as exact synonym for [is marker for](http://purl.obolibrary.org/obo/RO_0002607).

wdduncan:
* Fixes #860 
* added `altlabel` "indicates"
* @nataled rather than have long discussions about discussions about differences between narrow/broad/exact/related synonyms, how about we just use the parent annotation `altlabel`?


wdduncan:
Thanks @nataled  
Yes ... when I was making the PR I though altlabel may be better. 

Can you approve the PR?

nataled:
Actually, I take back what I said. When I see 'alternative' I tend to think it means 'exact'.

anitacaron (approved):

PR #886, opened 2026-07-22
Add three new property chains: has output > is small molecule regulator = directly regulates (and positive and negative)

### LINKED ISSUE #883: New property chain: has output > is small molecule regulator = directly regulates
Hi, 

We'd like to have a new property chain: has output (http://purl.obolibrary.org/obo/RO_0002234) > is small molecule regulator (http://purl.obolibrary.org/obo/RO_0012004) = directly regulates (http://purl.obolibrary.org/obo/RO_0002578)

I think the children do not need to be asserted, but I want to make sure that these inferences will also be made: 

has output > is small molecule activator (http://purl.obolibrary.org/obo/RO_0012005) = directly positively regulates (http://purl.obolibrary.org/obo/RO_0002629)  
has output > is small molecule inhibitor (http://purl.obolibrary.org/obo/RO_0012006) = directly negatively regulates (http://purl.obolibrary.org/obo/RO_0002630)

Thanks, Pascale

balhoff:
Fixes #883.


cmungall (approved):

cthoyt:
why is this not using the "directly regulates activity of" relations? typically, small molecules being activators or inhibitors is about their affect on the activity, not the amount

balhoff:
@cthoyt this chain is process to process, while 'directly regulates activity of' is material entity to material entity. Properties directly connecting the small molecule to a gene product could be created. However, I think 'directly regulates activity of' is in terms of 'enables', which RO doesn't use for small molecules.

cthoyt:
@balhoff thanks for the clarification

anitacaron (approved):

PR #894, opened 2026-09-04
Release September 2026

anitacaron:


balhoff (approved):
Thanks! Did not inspect every file, but looked at some the `.obo`s and they look right.

matentzn (approved):
Looks great!

